PR #112, opened 2018-03-08
Add new resized target option "w400"

babygoat:
This patch introduces new resized target option "w400" to improve
the hero image utilization.


nickhsine:
LGTM 💯

PR #118, opened 2018-04-12
Provide development environment by using docker-compose

nickhsine:
@babygoat


babygoat:
I found similar problem when starting the environment. I suggest we can try to add the "wait-for-it.sh" script to check the setup process of mysql.(e.g., wait-for-it.sh localhost:3306) In this way, we could make sure a user will always get a read-to-use set of environment.

babygoat:
LGTM 👍

PR #120, opened 2018-05-16
New endpoint for subscribing webpush notification

nickhsine:
@babygoat  

There are two new endpoints for creating/reading web push subscription.
One is **POST** method for creating a web push subscription,
another is **GET** method for reading a web push subscription.

### POST Body(either `application/JSON` or `x-www-form-urlencoded`): 
JSON example:
```
{
  "endpoint": "https://fcm.googleapis.com/fcm/send/f4Stnx6WC5s:APA91bFGo-JD8bDwezv1fx3RRyBVq6XxOkYIo8_7vCAJ3HFHLppKAV6GNmOIZLH0YeC2lM_Ifs9GkLK8Vi_8ASEYLBC1aU9nJy2rZSUfH7DE0AqIIbLrs93SdEdkwr5uL6skPMjJsMRT",
  "keys":"{\"p256dh\":\"BDmY8OGe-LfW0ENPIADvmdZMo3GfX2J2yqURpsDOn5tT8lQV-VVHyhRUgzjnmx_RRoobwdLULdBr26oULtLML3w\",\"auth\":\"P_AJ9QSqcgM-KJi_GRN3fQ\"}",
  "user_id": "1",
  "expirationTime": "1526457025"
}
```
### Response/ Status Code
**204** Created
**400** Bad request
**409** Duplicate record(depends on `endpoint` value)
**500** Internal server error

### Query Param: `endpoint`
Query example: `/v1/web-push/subscriptions?endpoint=https://fcm.googleapis.com/fcm/send/cmqyujQBXZU:APA91bHKpHph3rObZixXGpPHmX5dvd2FO5artSS86QeLK8VBaC91fcrJvqGfCe8n9YaDcamu6u07hx5i45sGUBs7hBRU39YIwduTNM3SIcr39frdl9KwlvVkeFiH8vqGPDo5JpSh92cn`
### Response/ Status Code
**200** Ok
**400** Bad request
**404** Record not found
**500** Internal server error

### TODO List
1) ~~Update README.md~~
2) ~~Functional test~~
3) Bind webpush subscription with user if needed in the future

### DIFF
--- a/controllers/subscription.go
+++ b/controllers/subscription.go
@@ -1,0 +1,95 @@
+package controllers
+
+import (
+	"crypto/md5"
+	"fmt"
+	"net/http"
+	"strconv"
+
+	//log "github.com/Sirupsen/logrus"
+	"github.com/gin-gonic/gin"
+	"twreporter.org/go-api/models"
+)
+
+// IsWebPushSubscribed - which handles the HTTP Get request,
+// and try to check if the web push subscription is existed or not
+func (mc *MembershipController) IsWebPushSubscribed(c *gin.Context) (int, gin.H, error) {
+	const errorWhere = "MembershipController.IsWebPushSubscribed"
+	var endpoint = c.Query("endpoint")
+	var err error
+	var hashEndpoint string
+	var wpSub models.WebPushSubscription
+
+	if endpoint == "" {
+		return http.StatusNotFound, gin.H{"status": "error", "message": "Fail to get a web push subscription since you do not provide endpoint in URL query param"}, nil
+	}
+
+	hashEndpoint = fmt.Sprintf("%x", md5.Sum([]byte(endpoint)))
+
+	if wpSub, err = mc.Storage.GetAWebPushSubscriptionByHashEndpoint(hashEndpoint); err != nil {
+		switch appErr := err.(type) {
+		case models.AppError:
+			return 0, gin.H{}, models.NewAppError(errorWhere, "Fail to get a web push subscription", appErr.Error(), appErr.StatusCode)
+		default:
+			return http.StatusInternalServerError, gin.H{"status": "error", "message": "Unknown Error Type. Fail to get a web push subscription"}, nil
+		}
+	}
+
+	return http.StatusOK, gin.H{"status": "success", "data": wpSub}, nil
+}
+
+// SubscribeWebPush - which handles the HTTP POST request,
+// and try to create a web push subscription record into the persistent database
+func (mc *MembershipController) SubscribeWebPush(c *gin.Context) (int, gin.H, error) {
+	// subscriptionBody is to store POST body
+	type subscriptionBody struct {
+		Endpoint       string `json:"endpoint" form:"endpoint" binding:"required"`
+		Keys           string `json:"keys" form:"keys" binding:"required"`
+		ExpirationTime string `json:"expirationTime" form:"expirationTime"`
+		UserID         string `json:"user_id" form:"user_id"`
+	}
+
+	const errorWhere = "MembershipController.SubscribeWebPush"
+	var err error
+	var sBody subscriptionBody
+	var expirationTime int64
+	var userID uint64
+	var wpSub models.WebPushSubscription
+
+	if err = c.Bind(&sBody); err != nil {
+		return http.StatusBadRequest, gin.H{"status": "fail", "data": subscriptionBody{
+			Endpoint:       "endpoint is required, and need to be a string",
+			Keys:           "keys is required, and need to be a string",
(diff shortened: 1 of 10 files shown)


controllers/subscription.go
```diff
@@ -0,0 +1,64 @@
...
+		wpSub.SetUserID(uint(userID))
+	}
+
+	if expirationTime, err = strconv.ParseInt(sBody.ExpirationTime, 10, 64); err == nil {
+		wpSub.SetExpirationTime(expirationTime)
+	}
+
+	if err = mc.Storage.CreateAWebpushSubscription(wpSub); err != nil {
+		appErr, _ := err.(models.AppError)
+		return 0, gin.H{}, models.NewAppError(errorWhere, "Creating a webpush subscription fails", err.Error(), appErr.StatusCode)
```
babygoat: http.StatusInternalServerError?

models/subscription.go
```diff
@@ -0,0 +1,32 @@
...
+
+// SetExpirationTime - set the pointer of expireTime into WebpushSubscription struct.
+// The reason why we set the pointer, not the value, into struct
+// is because we want to remain the NULL(nil) value if expireTime is not provided.
+func (wpSub *WebpushSubscription) SetExpirationTime(expireTime int64) {
+	_expireTime := time.Unix(expireTime, 0)
+	wpSub.ExpirationTime = &_expireTime
+}
+
+// SetUserID - set the pointer of userID into WebpushSubscription struct
```
babygoat: Why we prefer storing the pointer to actual user id value here?

controllers/subscription.go
nickhsine: It might be `http.StatusConflict`

models/subscription.go
nickhsine: If we don't store pointer here, the default value of uint will be `0`.

controllers/subscription.go
babygoat: appErr.StatusCode for the first return value?

models/subscription.go
babygoat: Any issue with the 0 default setting? It seems that the webpush_subscription table does not contain the unique key constraint on user_id field.

controllers/subscription.go
babygoat: Got it. Thanks!

models/subscription.go
nickhsine: Frankly speaking, it will not cause any problem if we set the value `0` into `userID` field.
However, I think `NULL` value be more accurate since we don't have a user whose ID is `0`.

nickhsine: But, if setting pointer into `userID` field is kind of hard to understanding, I can change the value to `0`

controllers/subscription.go
```diff
@@ -0,0 +1,91 @@
...
+	var endpoint = c.Query("endpoint")
+	var err error
+	var hashEndpoint string
+	var wpSub models.WebPushSubscription
+
+	if endpoint == "" {
+		return http.StatusBadRequest, gin.H{"status": "fail", "data": struct {
+			Query string
+		}{
+			Query: "query parameter `endpoint` should be provied",
```
babygoat: Typo. Provided?

membership_user.sql
```diff
@@ -171,3 +171,26 @@ CREATE TABLE `users_bookmarks` (
...
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `web_push_subscriptions` (
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `endpoint` varchar(2048) NOT NULL,
+  `hash_endpoint` char(32) NOT NULL,
+  `keys` varchar(200) NOT NULL,
+  `expiration_time` timestamp NULL DEFAULT NULL,
+  `user_id` int(10) unsigned NULL DEFAULT NULL,
+  `created_at` timestamp NULL DEFAULT NULL,
```
babygoat: Default to CURRENT_TIMESTAMP?

membership_user.sql
```diff
@@ -171,3 +171,26 @@ CREATE TABLE `users_bookmarks` (
...
+  `updated_at` timestamp NULL DEFAULT NULL,
```
babygoat: If we had expiration_time field for managing the endpoint update status, I assumed that should be enough for tracking. Is there any use case should be dealt with this separate field?

nickhsine: `expiration_time` is not always provided, it could be `NULL` most of the time.

nickhsine:
Adjustment according to review comments. 
If it is ok, I will merge this PR later.

babygoat:
LGTM 👍 👍 👍

PR #128, opened 2018-06-19
[Performance] Replace md5 hash function by crc32 on subscriptions table

nickhsine:
@babygoat


babygoat:
commit: a43090e seems to be redundant?

PR #129, opened 2018-06-19
[Performance] Replace md5 hash function by crc32 on subscriptions table

nickhsine:
@babygoat

### DIFF
--- a/membership_user.sql
+++ b/membership_user.sql
@@ -174,22 +174,23 @@ CREATE TABLE `users_bookmarks` (
 
 
 --
--- Table structure for table `web_push_subscription`
+-- Table structure for table `web_push_subs
 --
 
-DROP TABLE IF EXISTS `web_push_subscriptions`;
+DROP TABLE IF EXISTS `web_push_subs`;
 /*!40101 SET @saved_cs_client     = @@character_set_client */;
 /*!40101 SET character_set_client = utf8 */;
-CREATE TABLE `web_push_subscriptions` (
+CREATE TABLE `web_push_subs` (
   `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
-  `endpoint` varchar(2048) NOT NULL,
-  `hash_endpoint` char(32) NOT NULL,
+  `endpoint` varchar(500) NOT NULL,
+  `crc32_endpoint` int(10) unsigned NOT NULL,
   `keys` varchar(200) NOT NULL,
   `expiration_time` timestamp NULL DEFAULT NULL,
   `user_id` int(10) unsigned NULL DEFAULT NULL,
   `created_at` timestamp DEFAULT CURRENT_TIMESTAMP,
   `updated_at` timestamp DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
   PRIMARY KEY (`id`),
-  UNIQUE KEY `uix_web_push_subscriptions_hash_endpoint` (`hash_endpoint`)
+  UNIQUE KEY `uix_web_push_subs_endpoint` (`endpoint`),
+  KEY `idx_web_push_subs_crc32_endpoint` (`crc32_endpoint`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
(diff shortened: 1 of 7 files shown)


storage/subscription.go
```diff
@@ -1,26 +1,54 @@
...
+
+	// find if endpoint is existed or not
+	if err = tx.Find(&ws, "hash_endpoint = ? AND endpoint = ?", wpSub.HashEndpoint, wpSub.Endpoint).Error; err != nil {
+		// endpoint is not in the DB
+		if IsRecordNotFoundError(err) {
+			// create the subscription
+			err = tx.Create(&wpSub).Error
+			// no error occurs, commit the transaction
+			if err == nil {
+				return tx.Commit().Error
```
babygoat: What if an error occurs during the transaction commit? Is it possible for controller to handle the error?

nickhsine: The error will passed to controller, and controller will return internal server error.

storage/errors.go
```diff
@@ -24,14 +24,26 @@ var ErrDuplicateEntry uint16 = 1062
...
+	errStruct, ok := err.(*mysql.MySQLError)
+	if ok && errStruct.Number == ErrDuplicateEntry {
+		return true
+	}
+	return false
+}
+
 // NewStorageError return AppError with detailed information.
 // This method is mainly used to deal with MySQL error.
 func (g *GormStorage) NewStorageError(err error, where string, message string) error {
```
babygoat: When it comes to deal with the if-else-chain, I suggest use switch instead.

switch {
     case IsRecordNotFoundError(err):
              stmt1
     case IsDuplicateEntryError(err):
              stmt2
     default:
              default handling
}

storage/bookmark.go
```diff
@@ -83,7 +83,7 @@ func (g *GormStorage) GetBookmarksOfAUser(id string, limit, offset int) ([]model
 
 // CreateABookmarkOfAUser this func will create a bookmark and build the relationship between the bookmark and the user
 func (g *GormStorage) CreateABookmarkOfAUser(userID string, bookmark models.Bookmark) (models.Bookmark, error) {
-	var _bookmark models.Bookmark = bookmark
+	var _bookmark = bookmark
 	var err error
 	var user models.User
```
babygoat: Should below update operation wrapping within a transaction group?

nickhsine: Yeah, it should be wrapped in a transaction.
I will fix it in the next PR.

membership_user.sql
```diff
@@ -183,13 +183,13 @@ DROP TABLE IF EXISTS `web_push_subscriptions`;
...
-  `hash_endpoint` char(32) NOT NULL,
+  `hash_endpoint` int(10) unsigned NOT NULL,
   `keys` varchar(200) NOT NULL,
   `expiration_time` timestamp NULL DEFAULT NULL,
   `user_id` int(10) unsigned NULL DEFAULT NULL,
   `created_at` timestamp DEFAULT CURRENT_TIMESTAMP,
   `updated_at` timestamp DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
   PRIMARY KEY (`id`),
-  UNIQUE KEY `uix_web_push_subscriptions_hash_endpoint` (`hash_endpoint`)
+  KEY `uix_web_push_subscriptions_hash_endpoint` (`hash_endpoint`)
```
babygoat: Minor. If the index is no longer unique, I suggest the naming should be adapted accordingly.

babygoat:
Except for the minor suggestion, others look good to me now.

membership_user.sql
nickhsine: I will update this key name before merging this PR.

babygoat:
LGTM

models/subscription.go
```diff
@@ -29,3 +29,8 @@ func (wpSub *WebPushSubscription) SetExpirationTime(expireTime int64) {
 func (wpSub *WebPushSubscription) SetUserID(userID uint) {
 	wpSub.UserID = &userID
 }
+
+// set WebPushSubscription's table name ti be `web_push_subs`
```
babygoat: typo to?

PR #133, opened 2018-06-29
Fix circleci build failing. Install corresponding kubectl pkg version

nickhsine:
The latest version of kubectl(v1.11.0) breaks the circleci builds.
Before our containers upgrades to v1.11.0, we install kubectl clients according to cluster kubectl version.
@babygoat

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,54 +80,63 @@ jobs:
       - restore_cache:
           keys:
             - v1-build-{{ checksum ".circle-sha" }}
+      - run:
+          name: Connect to container cluster
+          command: |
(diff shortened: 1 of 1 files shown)


babygoat:
I suggest we can to get the gcloud kubernetes version so that we won't have to deal with the version upgrade of our clusters.

Maybe like this:
```
...
KUBECTL_VERSION = gcloud container clusters describe twreporter-website | grep version | tr -s ' ' | cut -d'-' -f1 | cut -d' ' -f3 
curl -LO "https://storage.googleapis.com/kubernetes-release/release/v$KUBERCTL_VERSION/bin/linux/amd64/kubectl"
...
```

nickhsine:
New commit according to review comments.

PR #135, opened 2018-07-18
Fix circle-ci config

babygoat:
This patch ports the circle-config fixes from twreporter-react repo
for following error.
* Missing CLUSTER_NAME environment variable in "kubectl install" step.
* Fetch the master node version of the cluster instead as there might
exist multiple versions within clusters.


nickhsine:
LGTM

PR #138, opened 2018-07-29
[code refactoring] use AppError precisely. avoid printing too many

nickhsine:
@babygoat


controllers/facebook-oauth.go
```diff
@@ -83,17 +84,26 @@ func (f *Facebook) BeginAuth(c *gin.Context) {
...
+			c.Redirect(http.StatusTemporaryRedirect, destination)
+		}
+	}()
+
+	destination = c.Query("destination")
 
 	// get user data from Facebook
-	fstring, err := f.GetRemoteUserData(c.Request, c.Writer)
-	if err != nil {
-		c.JSON(http.StatusUnauthorized, gin.H{"status": "error", "error": err.Error()})
```
babygoat: WHY can we replace StatusUnauthorized(401) with (TemporaryRedirect)307?

controllers/facebook-oauth.go
```diff
@@ -212,31 +212,27 @@ func (f *Facebook) GetRemoteUserData(r *http.Request, w http.ResponseWriter) (st
 	// get Facebook OAuth Token
 	state := r.FormValue("state")
 	if state != oauthStateString {
-		log.Warnf("controllers.oauth.facebook.getRemoteUserData. invalid oauth state, expected '%s', got '%s'\n", oauthStateString, state)
-		return "", models.NewAppError("OAuth state", "controllers.oauth.facebook", "Invalid oauth state", 500)
+		return "", models.NewAppError("Facebook.GetRemoteUserData", "invalid oauth state", fmt.Sprintf("invalid oauth state, expected '%s', actual '%s'\n", oauthStateString, state), 500)
```
babygoat: Minor. I suggest replace 500 with http.StatusInternalServerError

storage/bookmark.go
```diff
@@ -45,16 +46,17 @@ func (g *GormStorage) GetABookmarkOfAUser(userID string, bookmarkSlug string, bo
...
 
 	for _, ele := range bookmarks {
 		if ele.Slug == bookmarkSlug && ele.Host == bookmarkHost {
-			return ele, err
+			return ele, nil
 		}
 	}
 
-	return bookmark, models.NewAppError("GetABookmarkOfAUser", "Record not found", "storage.bookmark.bookmark_is_not_registered_by_user", 404)
+	return bookmark, models.NewAppError("GormStorage.GetABookmarkOfAUser", "Record not found", fmt.Sprintf("User %s does not have the bookmark whose slug is %s and host is %s", userID, bookmarkSlug, bookmarkHost), 404)
```
babygoat: Minor. http.StatusNotFound

utils/config.go
```diff
@@ -20,15 +20,15 @@ func LoadConfig(fileName string) error {
 
 	file, err := os.Open(fileName)
 	if err != nil {
-		appError := models.NewAppError("LoadConfig", "utils.config.load_conifg.open_file: ", err.Error(), 500)
+		appError := models.NewAppError("LoadConfig", "internal server error: fail to load config", err.Error(), 500)
```
babygoat: minor. http.StatusInternalServerError

storage/user.go
```diff
@@ -41,7 +42,7 @@ func (gs *GormStorage) GetOAuthData(aid sql.NullString, aType string) (models.OA
 	oac := models.OAuthAccount{}
 	err := gs.db.Where(&models.OAuthAccount{Type: aType, AId: aid}).Last(&oac).Error
 	if err != nil {
-		return oac, gs.NewStorageError(err, "GormStorage.GetOAuthData", "storage.user.get_o_auth_data.error")
+		return oac, gs.NewStorageError(err, "GormStorage.GetOAuthData", fmt.Sprint("get oauth account error"))
```
babygoat: Do we need fmt.Sprint?

controllers/bookmark.go
```diff
@@ -27,15 +26,11 @@ func (mc *MembershipController) GetBookmarksOfAUser(c *gin.Context) {
...
-			appErr = err.(models.AppError)
-			c.JSON(appErr.StatusCode, gin.H{"status": appErr.Message, "error": err.Error()})
-			return
+		if bookmark, err = mc.Storage.GetABookmarkOfAUser(userID, bookmarkSlug, host); err != nil {
+			return 0, gin.H{}, err
 		}
 
-		c.JSON(http.StatusOK, gin.H{"status": "ok", "record": bookmark})
-		return
+		return http.StatusOK, gin.H{"status": "ok", "record": bookmark}, nil
```
babygoat: For 2xx, 3xx.  I suggest we should use "success" or "ok" only in one repo.

controllers/bookmark.go
```diff
@@ -48,70 +43,54 @@ func (mc *MembershipController) GetBookmarksOfAUser(c *gin.Context) {
...
-	if err != nil {
-		appErr = err.(models.AppError)
-		c.JSON(appErr.StatusCode, gin.H{"status": appErr.Message, "error": err.Error()})
-		return
+	if bookmarks, total, err = mc.Storage.GetBookmarksOfAUser(userID, limit, offset); err != nil {
+		return 0, gin.H{}, err
 	}
 
-	c.JSON(http.StatusOK, gin.H{"status": "ok", "records": bookmarks, "meta": models.MetaOfResponse{
+	return http.StatusOK, gin.H{"status": "ok", "records": bookmarks, "meta": models.MetaOfResponse{
```
babygoat: Please follow {"status": "ok" or "success", "data": {}}

controllers/controller-factory.go
```diff
@@ -97,3 +97,13 @@ func NewControllerFactory() (*ControllerFactory, error) {
 
 	return cf, nil
 }
+
+// AppErrorTypeAssertion type assertion for *models.AppError
+func AppErrorTypeAssertion(err error) *models.AppError {
```
babygoat: If we only use this within the controller pkg scopes, we could use lowercase function name

controllers/bookmark.go
nickhsine: currently I would like to keep the old JSON data structure because we need to change FE code to apply this change.

babygoat: Sure. But we should leave a comment for the plan if we would like to format the response.

controllers/bookmark.go
nickhsine: Avoid effecting the current FE codes, I would like to apply this change later.

babygoat:
Why should we use _service, _registration instead of service and registration?

nickhsine:
duplicate of [140](https://github.com/twreporter/go-api/pull/140)

PR #140, opened 2018-07-31
[code refactoring] use AppError precisely. avoid printing too many

nickhsine:


babygoat:
Please update the test cases in `utils.config_test.go` as we plan to change the appError string representation.

PR #142, opened 2018-08-01
[feature] /v1/authors{?limit,offset,sort}[GET] endpoint for fetching authors

nickhsine:
@babygoat 

In order to let @YuCJ implement cron job to update algolia, we provide authors information by `/v1/authors{?limit,offset,sort}[GET]` endpoint.

### DIFF
--- a/controllers/author.go
+++ b/controllers/author.go
@@ -1,0 +1,47 @@
+package controllers
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"twreporter.org/go-api/models"
+)
+
+// GetAuthors receive HTTP GET method request, and return the authors.
+// `limit`, `offset` and `sort` are the url query params,
+// which define the rule we retrieve authors from storage.
+func (nc *NewsController) GetAuthors(c *gin.Context) (int, gin.H, error) {
+	const defaultLimit = 20
+	const defaultSort = "updatedAt"
+	var authors []models.FullAuthor
+	var err error
+	var total int
+
+	_, limit, offset, sort, _ := nc.GetQueryParam(c)
+
+	if limit == 0 {
+		limit = defaultLimit
+	}
+
+	if sort == "" {
+		sort = defaultSort
+	}
+
+	authors, total, err = nc.Storage.GetFullAuthors(limit, offset, sort)
+
+	if err != nil {
+		return 0, gin.H{}, err
+	}
+
+	return http.StatusOK, gin.H{
+		"status": "success",
+		"data": gin.H{
+			"records": authors,
+			"meta": models.MetaOfResponse{
+				Total:  total,
+				Offset: offset,
+				Limit:  limit,
+			},
+		},
+	}, nil
+}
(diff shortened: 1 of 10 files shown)


controllers/author.go
```diff
@@ -0,0 +1,45 @@
...
+// which define the rule we retrieve authors from storage.
+func (nc *NewsController) GetAuthors(c *gin.Context) (int, gin.H, error) {
+	var authors []models.FullAuthor
+	var err error
+	var total int
+
+	_, limit, offset, sort, _ := nc.GetQueryParam(c)
+
+	if limit == 0 {
+		limit = 20
```
babygoat: I suggest defining the constant for default value and then overwrites them if these parameters are provided. I believe it could improve the readibility.

controllers/author.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	if sort == "" {
+		sort = "updatedAt"
+	}
+
```
babygoat: I suggest still dealing with the offset even if its default value equals to `0`.

nickhsine: not sure what we should deal with `offset` if `offset == 0`.

storage/author.go
```diff
@@ -0,0 +1,66 @@
...
+
+	collection := m.db.DB(utils.Cfg.MongoDBSettings.DBName).C("contacts")
+	if total, err = collection.Count(); err != nil {
+		return authors, 0, models.NewAppError("MongoStorage.GetFullAuthors", "can not get total count of authors", err.Error(), http.StatusInternalServerError)
+	}
+
+	pipe := collection.Pipe(pipeline)
+
+	if err = pipe.All(&results); err != nil {
+		return authors, 0, models.NewAppError("MongoStorage.GetFullAuthors", "can not get authors from storage", err.Error(), http.StatusInternalServerError)
```
babygoat: Add log.Error(). Also, I suggest we should use NewStorageError(err, "MongoStorage.GetFullAuthors", "can not get authors from storage")  Avoiding directly calling NewAppError() in storage layer would make more sense to have NewStorageError().

storage/author.go
```diff
@@ -0,0 +1,66 @@
...
+		bson.M{"$sort": bson.M{sort: -1}},
+		bson.M{"$skip": offset},
+		bson.M{"$limit": limit},
+		bson.M{"$lookup": bson.M{"from": "images", "localField": "image", "foreignField": "_id", "as": "thumbnails"}},
+	}
+
+	collection := m.db.DB(utils.Cfg.MongoDBSettings.DBName).C("contacts")
+	if total, err = collection.Count(); err != nil {
+		return authors, 0, models.NewAppError("MongoStorage.GetFullAuthors", "can not get total count of authors", err.Error(), http.StatusInternalServerError)
+	}
```
babygoat: See below.

controllers/author.go
babygoat: Like 

```
_, tmpLimit, tmpOffset, tmpSort, _ = nc.GetQueryParameters(../)

if tmpSort != defaultSort {
   sort = tmpSort
}
```

However, I'm good to go with current implementation. This change would just be more intuitive for me.

PR #147, opened 2018-08-03
Periodic donation

babygoat:


controllers/donation.go
```diff
@@ -304,7 +402,35 @@ func buildPrimeSuccessRecord(resp tapPayTransactionResp) models.PayByPrimeDonati
...
+
+	m.CardInfoBinCode = &resp.CardInfo.BinCode
+	m.CardInfoLastFour = &resp.CardInfo.LastFour
+	m.CardInfoIssuer = &resp.CardInfo.Issuer
+	m.CardInfoFunding = &resp.CardInfo.Funding
+	m.CardInfoType = &resp.CardInfo.Type
+	m.CardInfoLevel = &resp.CardInfo.Level
+	m.CardInfoCountry = &resp.CardInfo.Country
+	m.CardInfoCountryCode = &resp.CardInfo.CountryCode
+	m.CardInfoExpiryDate = &resp.CardInfo.ExpiryDate
```
nickhsine: Are those pointer values `nil` if tappay response does not contain `card_info`?

controllers/donation.go
```diff
@@ -240,6 +322,22 @@ func buildClientResp(payMethod string, req tapPayPrimeReq, resp tapPayTransactio
...
 
+func buildDraftPeriodicDonation(userID uint, req clientReq) models.PeriodicDonation {
+	m := models.PeriodicDonation{}
+
+	m.CardholderEmail = req.Cardholder.Email
+	m.CardholderPhoneNumber = &req.Cardholder.PhoneNumber
+	m.CardholderName = &req.Cardholder.Name
+	m.CardholderZipCode = &req.Cardholder.ZipCode
+	m.CardholderAddress = &req.Cardholder.Address
+	m.CardholderNationalID = &req.Cardholder.NationalID
```
nickhsine: Are those pointer values `nil` if `req` does not contain `card_holder`?

controllers/donation.go
babygoat: Nope. It will be empty string `""`

controllers/donation.go
babygoat: Nope. It will be empty string `""`

babygoat: Currently, I cannot foresee any use case required to differentiate `NULL` from `""`.  I will replace all the fields with their underlying type.(i.e., remove pointer)

babygoat: Currently, I cannot foresee any use case required to differentiate `NULL` from `""`. I will replace all the (*string) fields with their underlying type.(i.e., remove pointer)

models/donation.go
```diff
@@ -96,20 +96,20 @@ type PeriodicDonation struct {
...
 	CardInfoFunding       *uint      `gorm:"type:tinyint" json:"card_info_funding"`
 	CardInfoType          *uint      `gorm:"type:tinyint" json:"card_info_type"`
-	CardInfoLevel         *string    `gorm:"type:varchar(10)" json:"card_info_level"`
-	CardInfoCountry       *string    `gorm:"type:varchar(30)" json:"card_info_country"`
-	CardInfoCountryCode   *string    `gorm:"type:varchar(10)" json:"card_info_country_code"`
-	CardInfoExpiryDate    *string    `gorm:"type:varchar(6);" json:"card_info_expiry_date"`
+	CardInfoLevel         string     `gorm:"type:varchar(10)" json:"card_info_level"`
+	CardInfoCountry       string     `gorm:"type:varchar(30)" json:"card_info_country"`
+	CardInfoCountryCode   string     `gorm:"type:varchar(10)" json:"card_info_country_code"`
+	CardInfoExpiryDate    string     `gorm:"type:varchar(6);" json:"card_info_expiry_date"`
```
nickhsine: This change is kind of confusing.
So, if the tappay response is not containing `card_info`, will we store empty string `''` for `string` fields, and `nil` for `*uint` fields?

babygoat: Yup, as I cannot see any difference between `null` and empty string `""`. Though, the raw record might be kind of tricky.

babygoat: Fixed.

PR #151, opened 2018-08-16
Fix missing field in token draft record

babygoat:
This patch sets the missing field `amount` when creating the draft token
record.


nickhsine:
LGTM

PR #152, opened 2018-08-21
Fix incorrect parsing of timestamp

babygoat:
This patch adds the fix of incorrect parsing of timestamp with unit
millisecond. Previous work treats the millisecond as nanosecond.


nickhsine:
LGTM

babygoat:
Clean rebase.

PR #153, opened 2018-08-24
Update transaction status field

babygoat:
This patch splits the third_party_status into tappay_api_status and
tappay_record_status to stores different types of status code. Also,
change their data types to pointer so that we could record status `0`.


nickhsine:
LGTM

babygoat:
Clean rebase.

PR #159, opened 2018-09-09
update config.go and oauth.go

nickhsine:
Update configs/config.go. change defaulthost.app.host and
default.app.domain to `localhost`
Update controllers/oauth.go. Use `https://www.twreporte.org/` as default destination

@babygoat

### DIFF
--- a/controllers/oauth.go
+++ b/controllers/oauth.go
@@ -24,14 +24,14 @@ import (
 const defaultDestination = "https://www.twreporter.org/"
 
 type basicInfo struct {
-	AId    models.NullString `json:"id"`
 	Email  models.NullString `json:"email"`
 	Name   models.NullString `json:"name"`
-	Gender models.NullString `json:"gender"`
+	Gender string            `json:"gender"`
 }
 
 type facebookOauthInfoRaw struct {
 	basicInfo
+	AId        models.NullString `json:"id"`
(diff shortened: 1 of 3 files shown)


routers/router.go
```diff
@@ -116,7 +116,7 @@ func SetupRouter(cf *controllers.ControllerFactory) *gin.Engine {
 	session := cf.GetMgoSession()
 	c := session.DB("go-api").C("sessions")
 	store := mongo.NewStore(c, maxAge, true, []byte("secret"))
-	v2AuthGroup.Use(sessions.Sessions("oauth-session", store))
+	v2AuthGroup.Use(sessions.Sessions("go-api-session", store))
```
babygoat: If we would like to rename the session-name to `go-api-session`, I suggest adding the session from v2Group.

controllers/oauth.go
```diff
@@ -24,14 +24,14 @@ import (
 const defaultDestination = "https://www.twreporter.org/"
 
 type basicInfo struct {
-	AId    models.NullString `json:"id"`
 	Email  models.NullString `json:"email"`
 	Name   models.NullString `json:"name"`
-	Gender models.NullString `json:"gender"`
+	Gender string            `json:"gender"`
```
babygoat: Why change this?

PR #160, opened 2018-09-10
Adapt tappay request fields

babygoat:
This patch changes the tappay request fields `cardholder.name` and
`cardholder.phonenumber` from presenting if provided by client to always
sending these fields.


nickhsine:
LGTM

PR #161, opened 2018-09-11
Change session cookie domain

babygoat:
This patch changes the session cookies domain to `.twreporter.org`
rather than default setting (HOST header in default request)

### DIFF
--- a/routers/router.go
+++ b/routers/router.go
@@ -13,7 +13,9 @@ import (
 	"twreporter.org/go-api/models"
 )
 
-const maxAge = 3600
+const (
+	maxAge = 3600
+)
 
 type wrappedFn func(c *gin.Context) (int, gin.H, error)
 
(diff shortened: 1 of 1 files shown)


routers/router.go
```diff
@@ -116,6 +120,10 @@ func SetupRouter(cf *controllers.ControllerFactory) *gin.Engine {
 	session := cf.GetMgoSession()
 	c := session.DB("go-api").C("sessions")
 	store := mongo.NewStore(c, maxAge, true, []byte("secret"))
+	store.Options(sessions.Options{
+		Domain: defaultDomain,
+		MaxAge: maxAge,
```
nickhsine: `HttpOnly:  globals.Conf.Environment != "development"`?

nickhsine: `Secure:  globals.Conf.Environment != "development"`?

nickhsine:
LGTM

PR #173, opened 2018-09-27
 use null library for dealing with nullable SQL and JSON values

nickhsine:
@babygoat

### DIFF
--- a/controllers/facebook-oauth.go
+++ b/controllers/facebook-oauth.go
@@ -20,6 +20,7 @@ import (
 
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/facebook"
+	"gopkg.in/guregu/null.v3"
 )
 
 const oauthState = "twreporter-state"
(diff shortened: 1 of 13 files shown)


controllers/facebook-oauth.go
```diff
@@ -100,13 +101,13 @@ func (f *Facebook) Authenticate(c *gin.Context) {
 	// decode user data returned by Facebook
 	remoteOAuth = models.OAuthAccount{
 		Type:      constants.Facebook,
-		AId:       models.NewNullString(gjson.Get(fstring, "id").Str),
-		Email:     models.NewNullString(gjson.Get(fstring, "email").Str),
-		Name:      models.NewNullString(gjson.Get(fstring, "name").Str),
-		FirstName: models.NewNullString(gjson.Get(fstring, "first_name").Str),
-		LastName:  models.NewNullString(gjson.Get(fstring, "last_name").Str),
+		AId:       null.StringFrom(gjson.Get(fstring, "id").Str),
```
babygoat: Should we use utils.ToNullString to wrap up the edge case if the any of these requested fields is empty string `""`?

nickhsine: if it is empty string,  we should store empty string.

nickhsine: unless the OAuth API does not reply these fields.

PR #174, opened 2018-10-01
Patch donations

nickhsine:
@babygoat


models/donation.go
```diff
@@ -7,44 +7,45 @@ import (
...
-	AuthCode                 string      `gorm:"type:varchar(6);not null" json:"auth_code"`
-	Amount                   uint        `gorm:"not null" json:"amount"`
-	Currency                 string      `gorm:"type:varchar(3);default:'TWD';not null" json:"currency"`
-	OrderNumber              string      `gorm:"type:varchar(50);not null" json:"order_number"`
 	Acquirer                 string      `gorm:"type:varchar(50);not null" json:"acquirer"`
-	TransactionTime          null.Time   `json:"transaction_time"`
-	BankTransactionStartTime null.Time   `json:"bank_transaction_start_time"`
-	BankTransactionEndTime   null.Time   `json:"bank_transaction_end_time"`
+	Amount                   uint        `gorm:"not null" json:"amount"`
+	AuthCode                 string      `gorm:"type:varchar(6);not null" json:"auth_code"`
```
babygoat: If we want to make the model members declared in alphabetic order, I suggest this should apply to the (to-change) file scope.

storage/membership.go
```diff
@@ -82,3 +76,87 @@ func (gs *GormStorage) Close() error {
...
+}
+
+// Update method of MembershipStorage interface
+func (gs *GormStorage) Update(m interface{}) error {
+	var err error
+	var errWhere string = "GormStorage.Update"
+
+	// caution:
+	// batch updates if primary key of m is zero value
+	err = gs.db.Model(m).Updates(m).Error
```
babygoat: I'm not favor of this design from gorm in case of carelessness. I suggest how about we change it to 

```
Update(id uint, m interface{}, mType Type) {
...
model := m.(mType)
model.ID = id
...
}
```

storage/membership.go
```diff
@@ -82,3 +76,87 @@ func (gs *GormStorage) Close() error {
...
+	if err != nil {
+		log.Error(err.Error())
+		return gs.NewStorageError(err, errWhere, fmt.Sprintf("can not update the record(%#v)", m))
+	}
+
+	return nil
+}
+
+// UpdateByConditions method of MembershipStorage interface
+func (gs *GormStorage) UpdateByConditions(cond map[string]interface{}, m interface{}) error {
```
babygoat: See above.

storage/membership.go
nickhsine: after discussion, I will remove `Update` function and only remain `UpdateByConditions` function.

babygoat:
LGTM

PR #183, opened 2018-10-11
endpoints for retrieving a donation of a user

nickhsine:
add three endpoints, including
- `/v1/users/:userID/periodic-donations/:id`, for retrieving a periodic donation according to its `id`
- `/v1/users/:userID/donations/prime/:id`, for retrieving a prime donation according to its `id`
- `/v1/users/:userID/donations/others/:id`, for retrieving a pay by other method donation according to its `id`

**TODO**
`/v1/users/:userID/donations/token/:id`, for retrieving a card token donation according to its `id`

### DIFF
--- a/controllers/donation.go
+++ b/controllers/donation.go
@@ -39,15 +39,18 @@ type (
 	}
 
 	clientResp struct {
-		PeriodicID  null.Int          `json:"periodic_id"`
-		PayMethod   string            `json:"pay_method"`
-		CardInfo    models.CardInfo   `json:"card_info"`
-		Cardholder  models.Cardholder `json:"cardholder"`
 		Amount      uint              `json:"amount"`
+		CardInfo    models.CardInfo   `json:"card_info,omitempty"`
+		Cardholder  models.Cardholder `json:"cardholder,omitempty"`
 		Currency    string            `json:"currency"`
-		Details     string            `json:"details"`
-		OrderNumber string            `json:"order_number"`
+		Details     string            `json:"details,omitempty"`
 		ID          uint              `json:"id"`
+		Notes       string            `json:"notes"`
+		OrderNumber string            `json:"order_number,omitempty"`
+		PayMethod   string            `json:"pay_method"`
+		PeriodicID  uint              `json:"periodic_id,omitempty"`
+		SendReceipt string            `json:"send_receipt"`
+		ToFeedback  bool              `json:"to_feedback"`
 	}
 
 	bankTransactionTime struct {
@@ -108,6 +111,7 @@ type (
 		Details     null.String `json:"details"`
 		Name        null.String `json:"name"`
 		NationalID  null.String `json:"national_id"`
+		Notes       null.String `json:"notes"`
 		PhoneNumber null.String `json:"phone_number"`
 		SendReceipt null.String `json:"send_receipt"`
 		ToFeedback  null.Bool   `json:"to_feedback"`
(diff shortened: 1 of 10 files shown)


controllers/donation.go
```diff
@@ -39,14 +39,14 @@ type (
 	}
 
 	clientResp struct {
-		PeriodicID  null.Int          `json:"periodic_id"`
+		PeriodicID  uint              `json:"periodic_id,omitempty"`
```
babygoat: Isn't this a required field?

controllers/donation.go
```diff
@@ -39,14 +39,14 @@ type (
...
-		CardInfo    models.CardInfo   `json:"card_info"`
-		Cardholder  models.Cardholder `json:"cardholder"`
+		CardInfo    models.CardInfo   `json:"card_info,omitempty"`
+		Cardholder  models.Cardholder `json:"cardholder,omitempty"`
 		Amount      uint              `json:"amount"`
 		Currency    string            `json:"currency"`
-		Details     string            `json:"details"`
-		OrderNumber string            `json:"order_number"`
+		Details     string            `json:"details,omitempty"`
+		OrderNumber string            `json:"order_number,omitempty"`
```
babygoat: see above

controllers/donation.go
```diff
@@ -115,6 +115,69 @@ type (
 	}
 )
 
+func (cr *clientResp) BuildFromPeriodicDonationModel(pd models.PeriodicDonation) {
+	cr.CardInfo = models.CardInfo{
```
babygoat: use `copier` package.

controllers/donation.go
```diff
@@ -115,6 +115,69 @@ type (
...
+		Email:       pd.CardholderEmail,
+		Name:        pd.CardholderName,
+		NationalID:  pd.CardholderNationalID,
+		PhoneNumber: pd.CardholderPhoneNumber,
+		ZipCode:     pd.CardholderZipCode,
+	}
+	cr.Amount = pd.Amount
+	cr.Currency = pd.Currency
+	cr.PayMethod = defaultPeriodicPayMethod
+	cr.ID = pd.ID
```
babygoat: use `copier` package?

controllers/donation.go
```diff
@@ -115,6 +115,69 @@ type (
...
+		NationalID:  pd.CardholderNationalID,
+		PhoneNumber: pd.CardholderPhoneNumber,
+		ZipCode:     pd.CardholderZipCode,
+	}
+	cr.Amount = pd.Amount
+	cr.Currency = pd.Currency
+	cr.Details = pd.Details
+	cr.OrderNumber = pd.OrderNumber
+	cr.PayMethod = pd.PayMethod
+	cr.ID = pd.ID
```
babygoat: see above

controllers/donation.go
```diff
@@ -115,6 +115,69 @@ type (
...
+}
+
+func (cr *clientResp) BuildFromOtherMethodDonationModel(d models.PayByOtherMethodDonation) {
+	cr.Amount = d.Amount
```
babygoat: see above.

routers/router.go
```diff
@@ -87,11 +87,29 @@ func SetupRouter(cf *controllers.ControllerFactory) *gin.Engine {
...
+	v1Group.GET("/users/:userID/periodic-donations/:id", middlewares.CheckJWT(), middlewares.ValidateUserID(), ginResponseWrapper(func(c *gin.Context) (int, gin.H, error) {
+		return mc.GetADonationOfAUser(c, globals.PeriodicDonationType)
+	}))
 	v1Group.POST("/users/:userID/donations/:pay_method", middlewares.CheckJWT(), middlewares.ValidateUserID(), ginResponseWrapper(mc.CreateADonationOfAUser))
 	v1Group.PATCH("/users/:userID/donations/prime/:id", middlewares.CheckJWT(), middlewares.ValidateUserID(), ginResponseWrapper(func(c *gin.Context) (int, gin.H, error) {
 		return mc.PatchADonationOfAUser(c, globals.PrimeDonaitionType)
 	}))
 	// v1Group.GET("/users/:userID/donations", middlewares.CheckJWT(), middlewares.ValidateUserID(), ginResponseWrapper(mc.GetDonationsOfAUser))
+	// one-time donation including credit_card, line pay, apple pay, google pay and samsung pay
+	v1Group.GET("/users/:userID/donations/prime/:id", middlewares.CheckJWT(), middlewares.ValidateUserID(), ginResponseWrapper(func(c *gin.Context) (int, gin.H, error) {
```
babygoat: The routing path seems to be ambiguous.  It seems to me that ad-hoc recordID for a user makes the semantic of url vague.  I suggest consolidating the listing endpoints into the same category (e.g, /v1/donations/:ID) and authorizing the request by the `id_token`. (user_id is a good candidate)

controllers/donation.go
nickhsine: it is a required field only if the `clientResp` is for card token donation.

controllers/donation.go
nickhsine: `Details` and `OrderNumber` is not specified in `PeriodicDonation`.

routers/router.go
nickhsine: `/v1/donations/:id` is ambiguous as well.
I will adopt `/v1/donations/prime/:id`.

nickhsine: About the endpoints, I will change `GET` method in this PR. 
`PATCH` and `POST` will be fixed in the next PR.

nickhsine: After discussion, 
the endpoints will be 
`/v1/periodic-donations/:recordID?user_id=userID`
`/v1/donations/prime/:recordID?user_id=userID`,
`/v1/donations/token/:recordID?user_id=userID`
`/v1/donations/others/:recordID?user_id=userID`.

nickhsine:
@babygoat use copier and  router path change.

membership_user.sql
```diff
@@ -244,6 +244,7 @@ CREATE TABLE `pay_by_prime_donations` (
   `card_info_country` varchar(30) DEFAULT NULL, 
   `card_info_country_code` varchar(10) DEFAULT NULL, 
   `card_info_expiry_date` varchar(6) DEFAULT NULL, 
+  `notes` varchar(100) DEFAULT NULL,
```
babygoat: Should we reserve the size for around 100 Chinese word or so?

nickhsine: sure, let me increase it to `256`?

membership_user.sql
```diff
@@ -322,6 +323,7 @@ CREATE TABLE `periodic_donations` (
...
```
babygoat: see above.

membership_user.sql
babygoat: Depends on the placeholder on the form of frontend

nickhsine: I prefer to design this field on the backend first, and notice frontend what size of this field we support.

babygoat: Then, reserve the size for 256 and put the 80-word constraint on frontend

nickhsine: by the way, I just check the default collation of utf8.
The default collation is `utf8_general_ci`, so `varchar(100)` is for 100 Chinese words tho.

nickhsine:
@babygoat I will merge this PR if you think it is ok.

membership_user.sql
babygoat: OK, I misunderstood the size which stands for character but not byte. BTW, collation is for the sorting order of the record which is encoded in the corresponding record. (utf-8 can have utf8_general_ci or utf8_encode_ci) I believe this has nothing to do with the size.

nickhsine: oh, I misunderstood collation as well.

PR #188, opened 2018-10-18
Code refactoring according to code review comments.

nickhsine:
@babygoat


controllers/donation.go
```diff
@@ -199,22 +202,28 @@ var cardInfoTypes = map[int64]string{
...
+	// gin.Context.Bind does not support to bind `JSON` body multiple times
+	// the alternative is to use gin.Context.ShouldBindBodyWith function to bind
+	if err = c.ShouldBindBodyWith(reqBody, binding.JSON); nil != err {
+
+		// bind other format rather than JSON
+		if err = c.Bind(reqBody); nil != err {
+			failData := gin.H{}
+
+			switch e := err.(type) {
+			case *json.UnmarshalTypeError:
```
babygoat: If we bind it in other format(e.g, XML), I believed it will not throw json error.

controllers/donation.go
```diff
@@ -497,6 +501,16 @@ func (mc *MembershipController) GetADonationOfAUser(c *gin.Context, donationType
 		return http.StatusInternalServerError, gin.H{"status": "error", "message": fmt.Sprintf("donation type %s is not supported", donationType)}, nil
 	}
 
+	if err != nil {
```
babygoat: Will this error handling be processed?

membership_user.sql
```diff
@@ -327,6 +330,8 @@ CREATE TABLE `periodic_donations` (
   PRIMARY KEY (`id`),
   KEY `idx_periodic_donations_status` (`status`),
   KEY `idx_periodic_donations_amount` (`amount`),
+  KEY `idx_periodic_donations_order_number` (`order_number`),
+  KEY `idx_periodic_donations_frequency` (`frequency`),
```
babygoat: Why building index on `frequency`? This field would only be set to two different value

CHANGELOG.md
```diff
@@ -16,12 +16,12 @@
 
 #### New Feature:
 ##### Donation endpoints
-  - /v1/users/:userID/periodic_donations endpoint with POST method
-  - /v1/users/:userID/:pay_method endpoint with POST method
+  - /v1/periodic_donations endpoint with POST method
+  - /v1/donations/:pay_method endpoint with POST method
```
babygoat: How about
```
/v1/donations/prime 
``
and client specify the payment type in request body and default to `credit_card`?

This would make the patch route  below more sense.

File: controllers/donation.go (hunk 2 above)
Comment (babygoat) [reply controllers/donation.go]: The error (err) might result from above switch statement. For each case, the err will be handled and return. Even if the exceptional default case is reached, it will also return. I was wondering if any error case is not covered in the above conditions.

Comment (nickhsine) [reply controllers/donation.go]: oops, my bad, I forget to remove the switch case error handling.

File: middlewares/jwt.go [hunk 5]
```diff
@@ -43,6 +44,39 @@ func ValidateUserID() gin.HandlerFunc {
...
 
+type reqBody struct {
+	UserID uint64 `json:"user_id" form:"user_id" binding:"required"`
+}
+
+func ValidateUserIDInReqBody() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		var body reqBody
+		var err error
+		user := c.Request.Context().Value("user")
```
babygoat: I think this refactor is a good chance to rename the  jwt property in the `user` field of the request context which was too implicit. Try to set the `UserProperty` field

```
var jwtMiddleware = jwtmiddleware.New(jwtmiddleware.Options{
         ...
         UserProperty: "jwt",  // field name in the request context which stores the jwt token after parsing.
})

```

CHANGELOG.md
```diff
@@ -1,15 +1,3 @@
...
-  - /v1/mail/send_activation (POST method)
-  - /v1/mail/send_success_donation (POST method)
-
-#### Code refactoring
-  - add controllers/mail.go to handle HTTP request/response
-  - replace utils/mail.go by services/mail.go
-  - use template/signin.tmpl to generate activation mail HTML
-  - use template/success-donation.tmpl to generate success donation mail HTML
-  - send HTTP POST request to mail endpoints after signin and donation
-
```
babygoat: In my opinion, as we roll up to the 3.0.1 already, we should bump the version (4.0.0) to address this change instead of roll out the version. What do u think?

doc/prime-donation.apib
```diff
@@ -0,0 +1,296 @@
...
+                "donator": {
+                    "email": "developer@twporter.org"
+                },
+                "currency": "TWD",
+                "details": "報導者單筆捐款",
+                "merchant_id": "twreporter_CTBC",
+                "pay_method": "credit_card",
+                "prime": "test_3a2fb2b7e892b914a03c95dd4dd5dc7970c908df67a49527c0a648b2bc9",
+                "user_id": 1,
+            }
```
babygoat: I suggest declaring the data schema in another data structure like below and describing all the possible fields so that the generated HTML will contain the format and the field status (required or not). 

```
### PrimeCreationBody
+ amount: 500 (required, number)
+ donator
        + email: developer@twreporter.org(required)
+ currency: TWD (optional)
```

doc/prime-donation.apib
```diff
@@ -0,0 +1,296 @@
...
+
++ Response 500 (application/json)
+
+    
+    + Body
+
+            {
+                "status": "error",
+                "message": "unknow error."
```
babygoat: Typo. Unknown error.

doc/periodic-donation.apib
```diff
@@ -0,0 +1,298 @@
...
+                },
+                "currency": "TWD",
+                "details": "報導者定期定額捐款",
+                "frequency": "monthly",
+                "notes": "第一次捐給報導者喔",
+                "order_number": "twreporter-153985253506653918900",
+                "send_receipt": "monthly",
+                "to_feedback": true
+            }
+
```
babygoat: See prime request data format

tests/controller_donations_test.go
```diff
@@ -274,25 +310,27 @@ func testCreateADonationRecord(t *testing.T, path string, isPeriodic bool, autho
...
+			Frequency:  frequency,
+			MerchantID: testMerchantID,
+			Prime:      testPrime,
+			UserID:     userID,
 		}
 
 		reqBodyInBytes, _ = json.Marshal(reqBody)
 		resp = serveHTTP("POST", path, string(reqBodyInBytes), "application/json", authorization)
 		resBodyInBytes, _ = ioutil.ReadAll(resp.Result().Body)
+		fmt.Print(string(resBodyInBytes))
```
babygoat: Is this a debug message?

doc/periodic-donation.apib
```diff
@@ -0,0 +1,298 @@
...
+                "donator": {
+                    "email": "developer@twporter.org"
+                },
+                "currency": "TWD",
+                "details": "報導者定期定額捐款",
+                "merchant_id": "twreporter_CTBC",
+                "prime": "test_3a2fb2b7e892b914a03c95dd4dd5dc7970c908df67a49527c0a648b2bc9",
+                "frequency": "monthly",
+                "user_id": 1,
+            }
```
babygoat: See prime donation creation request data format

doc/periodic-donation.apib
```diff
@@ -0,0 +1,298 @@
...
+                    "zip_code": "104",
+                    "phone_number": "+886912345678",
+                    "national_id": "A12345678"
+                },
+                "notes": "第一次捐款報導者喔",
+                "send_receipt": "yearly",
+                "to_feeback": "no",
+                "user_id": 1,
+            }
+
```
babygoat: See prime donation creation request data format

doc/prime-donation.apib
```diff
@@ -0,0 +1,296 @@
...
+                    "national_id": "A12345678"
+                },
+                "currency": "TWD",
+                "details": "報導者單筆捐款",
+                "frequency": "one_time",
+                "notes": "第一次捐給報導者喔",
+                "order_number": "twreporter-153985253506653918900",
+                "pay_method": "credit_card",
+                "send_receipt": "monthly"
+            }
```
babygoat: See prime donation creation request data format

middlewares/jwt.go
```diff
@@ -43,6 +44,38 @@ func ValidateUserID() gin.HandlerFunc {
 	}
 }
 
+type reqBody struct {
```
babygoat: If this type is only used in ValidateUserIDInReqBody(), I suggest moving the declaration to the function local scope.

tests/controller_donations_test.go
```diff
@@ -242,30 +279,29 @@ func testCreateADonationRecord(t *testing.T, path string, isPeriodic bool, autho
...
 	t.Run("StatusCode=StatusCreated", func(t *testing.T) {
 		reqBody = requestBody{
-			Prime:       testPrime,
-			Amount:      testAmount,
-			Currency:    testCurrency,
-			Details:     testDetails,
-			OrderNumber: testOrderNumber,
-			MerchantID:  testMerchantID,
-			Cardholder:  testCardholder,
+			Prime:      testPrime,
```
babygoat: Alphabetic order?

tests/controller_donations_test.go
```diff
@@ -398,40 +438,50 @@ func TestCreateAPeriodicDonation(t *testing.T) {
...
+		resp = serveHTTP("POST", path, `{"user_id":1000}`, "application/json", authorization)
 		assert.Equal(t, http.StatusForbidden, resp.Code)
 	})
 
 	// ==========================================
 	// Test Periodic Donation Creation
 	// =========================================
-	path = "/v1/users/1/periodic_donations"
-	testCreateADonationRecord(t, path, true, authorization)
+	path = "/v1/periodic-donations"
```
babygoat: move the assignment to top 

```
var path = "/v1/periodic-donations"
```
 
instead of assignment in each scope.

tests/controller_donations_test.go
```diff
@@ -398,40 +438,50 @@ func TestCreateAPeriodicDonation(t *testing.T) {
...
+			NationalID:  null.StringFrom(testNationalID),
+			PhoneNumber: null.StringFrom(testPhoneNumber),
+			ZipCode:     null.StringFrom(testZipCode),
 		},
 		MerchantID: testMerchantID,
+		UserID:     userID,
+		Details:    testDetails,
+	}
+
+	if frequency != "" {
```
babygoat: I suggest you should do the assignment directly within the requestBody initialization.

tests/controller_donations_test.go
```diff
@@ -468,20 +518,22 @@ func TestPatchAPeriodicDonation(t *testing.T) {
 	authorization := fmt.Sprintf("Bearer %s", jwt)
 
 	t.Run("StatusCode=StatusUnauthorized", func(t *testing.T) {
-		path := fmt.Sprintf("/v1/users/%d/periodic_donations/1", user.ID)
+		path := fmt.Sprintf("/v1/periodic-donations/%d", res.Data.ID)
```
babygoat: How about moving this statement on top instead of duplicated in each test case? Leave the path declaration in NotFound test case for overwriting in that scope.

tests/controller_donations_test.go
```diff
@@ -616,19 +685,30 @@ func TestGetAPrimeDonationOfAUser(t *testing.T) {
 		path := fmt.Sprintf("/v1/donations/prime/%d?user_id=%d", primeRes.Data.ID, user.ID)
 		resp := serveHTTPWithCookies("GET", path, "", "application/json", authorization, cookie)
 		respInBytes, _ := ioutil.ReadAll(resp.Result().Body)
+		fmt.Print(string(respInBytes))
```
babygoat: debug message?

tests/controller_donations_test.go
```diff
@@ -529,22 +588,26 @@ func TestPatchAPrimeDonation(t *testing.T) {
 	authorization := fmt.Sprintf("Bearer %s", jwt)
 
 	t.Run("StatusCode=StatusUnauthorized", func(t *testing.T) {
-		path := fmt.Sprintf("/v1/users/%d/donations/prime/1", user.ID)
+		path := fmt.Sprintf("/v1/donations/prime/%d", res.Data.ID)
```
babygoat: See TestPatchaPeriodicDonation()

tests/controller_donations_test.go
```diff
@@ -157,9 +162,10 @@ func testDonationDataValidation(t *testing.T, path string, userID uint, authoriz
 		// - Lack of Cardholder
 		// ===========================================
 		reqBody = requestBody{
-			Prime:  testPrime,
-			Amount: testAmount,
-			UserID: userID,
+			Prime:     testPrime,
+			Amount:    testAmount,
+			UserID:    userID,
+			PayMethod: payMethods[0],
```
babygoat: Minor. How about declare individual string for each pay method rather than referencing in slice for readability?  If for time concerns, I'm ok to go with current implementation.

nickhsine: will declare individual string for each pay method.

controllers/donation.go
```diff
@@ -28,27 +28,29 @@ import (
...
-		ResultUrl   linePayResultUrl  `json:"result_url" form:"result_url"`
+		Amount     uint              `json:"amount" form:"amount" binding:"required"`
+		Cardholder models.Cardholder `json:"donator" form:"donator" binding:"required,dive"`
+		Currency   string            `json:"currency" form:"currency"`
+		Details    string            `json:"details" form:"details"`
+		MerchantID string            `json:"merchant_id" form:"merchant_id"`
+		PayMethod  string            `json:"pay_method" form:"pay_method"`
+		Prime      string            `json:"prime" form:"prime" binding:"required"`
+		ResultUrl  linePayResultUrl  `json:"result_url" form:"result_url"`
+		Frequency  string            `json:"frequency"`
```
babygoat: alphabetic order?

controllers/donation.go
```diff
@@ -107,43 +100,181 @@ type (
...
 	}
 )
 
-func (cr *clientResp) BuildFromPeriodicDonationModel(pd models.PeriodicDonation) {
-	cardInfo := models.CardInfo{}
-	cardholder := models.Cardholder{}
+func (p *patchBody) BuildPeriodicDonation() models.PeriodicDonation {
+	m := new(models.PeriodicDonation)
+	m.Cardholder = p.Donator
+	m.Notes = p.Notes
```
babygoat: use copier?

controllers/donation.go
```diff
@@ -107,43 +100,181 @@ type (
...
+	m.SendReceipt = p.SendReceipt
+	m.UserID = p.UserID
+	return *m
 }
 
-func (cr *clientResp) BuildFromPrimeDonationModel(pd models.PayByPrimeDonation) {
-	cardInfo := models.CardInfo{}
-	cardholder := models.Cardholder{}
+func (req clientReq) BuildTapPayReq(orderNumber string) tapPayPrimeReq {
+	primeReq := new(tapPayPrimeReq)
```
babygoat: If we want to consolidate all the donation requests, I might suggest renaming the tapPayPrimeReq to tapPayTransactionReq instead so that `prime` in the url and in our implementation will not be ambiguous.

controllers/donation.go
```diff
@@ -247,24 +381,42 @@ func (mc *MembershipController) CreateAPeriodicDonationOfAUser(c *gin.Context) (
...
+	var err error
 	var reqBody clientReq
+	var tapPayResp tapPayTransactionResp
 
 	if failData, valid := bindRequestBody(c, &reqBody); valid == false {
 		return http.StatusBadRequest, gin.H{"status": "fail", "data": failData}, nil
 	}
 
-	userID, _ := strconv.ParseUint(c.Param("userID"), 10, strconv.IntSize)
+	if reqBody.Cardholder.Email == "" {
```
babygoat: This case will be handle in the gin validator.

controllers/donation.go
```diff
@@ -293,52 +444,58 @@ func (mc *MembershipController) CreateAPeriodicDonationOfAUser(c *gin.Context) (
...
 
-	updateRecord := buildTokenSuccessRecord(tapPayResp)
-	updatePeriodicDonation := buildSuccessPeriodicDonation(tapPayResp)
-	updateRecord.ID = draftRecord.ID
+	// append tappay response onto donation model
+	tapPayResp.AppendRespOnPerodicDonation(&draftPeriodicDonation)
+	tapPayResp.AppendRespOnTokenDonation(&draftRecord)
 
-	if err = mc.Storage.UpdatePeriodicAndCardTokenDonationInTRX(draftPeriodicDonation.ID, updatePeriodicDonation, updateRecord); nil != err {
+	if err = mc.Storage.UpdatePeriodicAndCardTokenDonationInTRX(draftPeriodicDonation.ID, draftPeriodicDonation, draftRecord); nil != err {
```
babygoat: I still get confused by the naming. If you prefer to reusing the variable, I might suggest removing the draft prefix.

controllers/donation.go
```diff
@@ -247,24 +381,42 @@ func (mc *MembershipController) CreateAPeriodicDonationOfAUser(c *gin.Context) (
...
-	tapPayReq := buildTapPayPrimeReq(periodic, defaultPeriodicPayMethod, reqBody)
+	// generate token donation order number
+	dOrderNumber := generateOrderNumber(token, getPayMethodID(payMethodCollections[0]))
+	// Build a draft card token donation record
+	draftRecord := reqBody.BuildTokenDraftRecord(dOrderNumber)
 
-	draftRecord := buildTokenDraftRecord(tapPayReq)
+	// Build Tappay prime request
+	tapPayReq := reqBody.BuildTapPayReq(dOrderNumber)
+	tapPayReq.Details = fmt.Sprintf("%s;%s", "第一筆定期定額捐款", tapPayReq.Details)
```
babygoat: declare the variable for the constant string.

controllers/donation.go
```diff
@@ -293,52 +444,58 @@ func (mc *MembershipController) CreateAPeriodicDonationOfAUser(c *gin.Context) (
...
+	payMethod := reqBody.PayMethod
+	if err = validatePayMethod(reqBody.PayMethod); nil != err {
+		return http.StatusBadRequest, gin.H{"status": "fail", "data": gin.H{"req.Body.pay_method": err.Error()}}, nil
 	}
 
-	var reqBody clientReq
-
-	if failData, valid := bindRequestBody(c, &reqBody); valid == false {
-		return http.StatusBadRequest, gin.H{"status": "fail", "data": failData}, nil
+	if reqBody.Cardholder.Email == "" {
```
babygoat: Handle by gin validator.

controllers/donation.go
```diff
@@ -293,52 +444,58 @@ func (mc *MembershipController) CreateAPeriodicDonationOfAUser(c *gin.Context) (
...
+		return http.StatusBadRequest, gin.H{"status": "fail", "data": gin.H{
+			"req.Body.donator.email": "donator email is not valid",
+		}}, nil
 	}
 
-	userID, _ := strconv.ParseUint(c.Param("userID"), 10, strconv.IntSize)
+	// generate token donation order number
```
babygoat: prime donation

controllers/donation.go
```diff
@@ -293,52 +444,58 @@ func (mc *MembershipController) CreateAPeriodicDonationOfAUser(c *gin.Context) (
...
+	dOrderNumber := generateOrderNumber(prime, getPayMethodID(payMethod))
+	// Build a draft card token donation record
```
babygoat: prime donation

doc/periodic-donation.apib
```diff
@@ -0,0 +1,286 @@
...
+                "status": "fail",
+                "data": {
+                    "prime": "prime(string) is required",
+                    "amount": "amount(number) is required",
+                    "donator": {
+                        "email": "email(string) is required",
+                    },
+                    "details": "details(string) is optional",
+                    "merchant_id": "merchant_id(string) is optional",
+                    "frequency": "frequency(string) is optional. only support 'monthly' and 'yearly'",
```
babygoat: default to 'monthly'.

doc/periodic-donation.apib
```diff
@@ -0,0 +1,286 @@
...
+    + Headers
+
+            Content-Type: application/json
+            Authorization: Bearer <jwt>
+            
+    + Attributes (object)
+        + amount: 500 (required, number)
+        + currency: TWD 
+        + details: 報導者定期定額捐款
+        + donator (object)
```
babygoat: required

controllers/donation.go
babygoat: Yup. That is the trade off. I'm second to do the assignment independently. However, the pattern should keep up throughout the code base.

middlewares/jwt.go
```diff
@@ -43,6 +44,36 @@ func ValidateUserID() gin.HandlerFunc {
 	}
 }
 
+func ValidateUserIDInReqBody() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		var body = struct {
+			UserID uint64 `json:"user_id" form:"user_id" binding:"required"`
+		}{}
+		var err error
+		user := c.Request.Context().Value("user")
```
babygoat: Leave a TODO comment here for memo.

tests/controller_donations_test.go
```diff
@@ -623,12 +721,22 @@ func TestGetAPrimeDonationOfAUser(t *testing.T) {
 		json.Unmarshal(respInBytes, &resBody)
 		assert.Equal(t, http.StatusOK, resp.Code)
 		assert.Equal(t, testAmount, resBody.Data.Amount)
+		assert.Equal(t, testDetails, resBody.Data.Details)
```
babygoat: Minor. I might suggest declaring the expectResBody and use for match.

tests/controller_donations_test.go
```diff
@@ -684,9 +792,22 @@ func TestGetAPeriodicDonationOfAUser(t *testing.T) {
...
 		assert.Equal(t, donatorEmail, resBody.Data.Cardholder.Email)
+		assert.Equal(t, testAddress, resBody.Data.Cardholder.Address.ValueOrZero())
```
babygoat: See above.

controllers/donation.go
nickhsine: Before struct embedding, the field assignments are too much, the `card_info.*`, `cardholder.*` and `tappayRes.*` are almost 20 field assignments. 

Currently, I think we don not need copier to  copy those fields.
Besides, I think `copier` could be used in other situation. But for this case(related to money), I think field assignments are pretty straightforward.

controllers/donation.go
```diff
@@ -107,43 +100,181 @@ type (
...
+	if req.Frequency != "" {
+		m.Frequency = req.Frequency
+	} else {
+		m.Frequency = monthlyFrequency
+	}
+
+	if req.Details != "" {
+		m.Details = req.Details
+	} else {
+		m.Details = "一般線上定期定額捐款"
```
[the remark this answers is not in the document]
nickhsine: for one time usage, I prefer not to give it a const variable.

controllers/donation.go
nickhsine: I prefer not to make one time usage string to a const.

controllers/donation.go
nickhsine: Even we consolidate all the donation requests from clients(we did it at the beginning), we only send pay by prime request to TapPay. Hence, I think this naming is quiet understandable for me.

controllers/donation.go
nickhsine: will remove draft prefix.

nickhsine:
I want to mention what I am aiming to do in this PR.
1. Use struct embedding to reduce the duplicate codes.
2. Use struct methods to replace package level private functions.
3. Use single model or as fewer structs as we could to build what we need and  to reduce the code complexity.

Thanks for your patient and detailed code review.

nickhsine:
If there are no other things to change, I will modify the codes according to your comments, and merge this PR onto staging for further testing.

controllers/donation.go
babygoat: Depends on the perspective, if we stand in the shoes of a community developer who is just  involved in, I believe this will get him confused. As for maintaining a public repo, I will take this into consideration during our development. What do you think?

controllers/donation.go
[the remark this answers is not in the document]
babygoat: Since the const variable might be declared in the file scope (before all the function) or from the begin of the function, this will make developers easier to find and change the value with good variable naming. However, this is just coding style preference. I'm fine to go with the magic constant.

controllers/donation.go
babygoat: See the constant discussion on #L192

controllers/donation.go
babygoat: The validator do not just empty string checking but also email format verification. If for the completed validation, this should be considered.

babygoat:
I proposed some feedbacks for you to do the final judgement. I'm fine to go on with current implementation for testing.

controllers/donation.go
nickhsine: sure, will change it to `tapPayTransactionReq`.

controllers/donation.go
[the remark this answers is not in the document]
nickhsine: I think this code style convention could be aligned, I will use a const in the function scope.
By the way, if we change this one time string to a const, do we need to avoid using `:=`?
Without IDE helpers, using `:=` makes the codes hard to trace.
How do you think?
We may have further discussion about coding convention in person.

controllers/donation.go
nickhsine: will use other mail validation function to validate this field.

PR #189, opened 2018-10-19
hot fix(bug). remove middlewares.jwt.ValidateIDToken

nickhsine:
Panic will arise when type assertion.

Since the following code didn't work,
`user.(*jwt.Token).Claims.(utils.IDTokenJWTClaims)`,
we may need to find another way to check IDTokenJWTClaims's validation.

However, if we change the code back to previous commit, like below,
`user.(*jwt.Token).Claims.(jwt.MapClaims)`,
it does the same thing(parse JWT and check its validation)
what `CheckJWT` already does.

Hence, remove ValidateIDToken function before we find another way
to check IDTokenJWTClaims's validation.

@babygoat

### DIFF
--- a/middlewares/jwt.go
+++ b/middlewares/jwt.go
@@ -79,22 +79,6 @@ func ValidateAuthentication() gin.HandlerFunc {
 	}
 }
 
-func ValidateIDToken() gin.HandlerFunc {
-	return func(c *gin.Context) {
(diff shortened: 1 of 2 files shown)


nickhsine:
Merge it first. 
If have any concern, we could discuss in this PR.

nickhsine:
By the way, we may need to add function test on these login endpoints.

PR #197, opened 2018-10-25
update table schema character set and mysql connection charset

nickhsine:
### Problem
MySQL utf8 character set is not containing emoji utf8 unicodes.
They will by errors when inserting/updating emoji utf8 unicodes into databases with utf8 charset.

### Solution
Set MySQL connections and database charset to utf8mb4


babygoat:
Don't forget to update the connection in `pre_test_environment_setup.go`.

PR #202, opened 2018-11-20
Improvement

nickhsine:


routers/router.go
```diff
@@ -91,15 +91,15 @@ func SetupRouter(cf *controllers.ControllerFactory) *gin.Engine {
 	v1Group.DELETE("/users/:userID/bookmarks/:bookmarkID", middlewares.CheckJWT(), middlewares.ValidateUserID(), middlewares.SetCacheControl("no-store"), ginResponseWrapper(mc.DeleteABookmarkOfAUser))
 
 	// endpoints for donation
-	v1Group.POST("/periodic-donations", middlewares.CheckJWT(), middlewares.ValidateUserIDInReqBody(), middlewares.SetCacheControl("no-store"), ginResponseWrapper(mc.CreateAPeriodicDonationOfAUser))
-	v1Group.PATCH("/periodic-donations/:id", middlewares.CheckJWT(), middlewares.ValidateUserIDInReqBody(), middlewares.SetCacheControl("no-store"), ginResponseWrapper(func(c *gin.Context) (int, gin.H, error) {
+	v1Group.POST("/periodic-donations", middlewares.ValidateAuthentication(), middlewares.CheckJWT(), middlewares.ValidateUserIDInReqBody(), middlewares.SetCacheControl("no-store"), ginResponseWrapper(mc.CreateAPeriodicDonationOfAUser))
```
babygoat: I suggest renaming the CheckJWT() method to ValidateAuthorization() for distinguishing between the jwt usage in cookies and header.

doc/periodic-donation.apib
```diff
@@ -105,6 +106,7 @@ The states *id* and *order_number* are assigned by the TWReporter Go API at the
     + Headers
 
             Content-Type: application/json
```
babygoat: Do we need the `Content-Type` header?

doc/periodic-donation.apib
```diff
@@ -182,6 +184,7 @@ The states *id* and *order_number* are assigned by the TWReporter Go API at the
...
```
babygoat: Same above.

nickhsine:
Updated.

doc/periodic-donation.apib
babygoat: Yup, we should accept the header for negotiation. After digging into gin-contrib/cors I found that the `Content-Type` header is within the default header allowance so the changes should be fine.

routers/router.go
```diff
@@ -40,19 +40,27 @@ func SetupRouter(cf *controllers.ControllerFactory) *gin.Engine {
...
+		default:
+			// omit intentionally
+			break
+		}
 	}
 
 	config.AddAllowHeaders("Authorization")
-	config.AddAllowHeaders("ContentType")
 	config.AddAllowMethods("DELETE")
 	config.AddAllowMethods("PATCH")
```
babygoat: The `DELETE` and `PATCH` methods are allowed in the DefaultConfig(). So the above two lines are redundant.

nickhsine: [current version](https://github.com/gin-contrib/cors/blob/567de191692713543513692ecba9f6ca08cd660a/cors.go#L80) we use does not contain `DELETE` and `PATCH`.

babygoat:
LGTM. Please rebase the patches before merge.

nickhsine:
@babygoat one more commit to review, thanks.

middlewares/jwt.go
```diff
@@ -15,32 +15,50 @@ import (
...
+		var userProperty interface{}
+		var claims jwt.MapClaims
+
+		if err = jwtMiddleware.CheckJWT(c.Writer, c.Request); err != nil {
+			c.AbortWithStatus(http.StatusUnauthorized)
+			return
+		}
+
+		userProperty = c.Request.Context().Value(authUserProperty)
+		claims = userProperty.(*jwt.Token).Claims.(jwt.MapClaims)
```
babygoat: According to jwt spec, audience could either be a single case-sensitive string or an array of case-sensitive string. (rfc7519#sec-4.1.3)[https://tools.ietf.org/html/rfc7519#section-4.1.3]. Also, it seems to be a will-adopted feature in jwt-go. See (discussion)[https://github.com/dgrijalva/jwt-go/issues/184]. I suggest defining a wrapper function for this to support audience validation against []string. Something like below
```
func audienceValidationWrapper(claims jwt.MapClaims,audiences []string, required bool) bool{
         isValid := true
         for _, aud := range audiences {
               if !claims.VerifyAudience(aud, required) {
                 isValid = false 
                 break
                }
         }
         return isValid
}

...

if !audienceValidationWrapper(cliams, globals.Conf.App.JwtAudience, verifyRequired) || .. {
   ...
} 
```

nickhsine: Since [this issue](https://github.com/dgrijalva/jwt-go/issues/184) has not been implemented yet, `jtw.StandardClaims.Audience` and `jtw.MapClaims["aud"]` is currently `string` type, rather than `[]string` type.

Therefore,  for now,  it is hard to for-loop the `jwt.aud` to check if there is any string matches `globals.Conf.App.JwtAudience`.

I will leave `TODO` here  instead.
How do you think?

nickhsine:
If the latest commit is ok for now, I would like to merge this PR.

middlewares/jwt.go
```diff
@@ -108,40 +129,8 @@ func ValidateAuthentication() gin.HandlerFunc {
...
-	var whiteList = []string{"nickhsine@twreporter.org", "han@twreporter.org", "yucj@twreporter.org", "developer@twreporter.org"}
-	return func(c *gin.Context) {
-		user := c.Request.Context().Value("user")
-		userIDClaim := user.(*jwt.Token).Claims.(jwt.MapClaims)["email"]
-
-		for _, v := range whiteList {
-			if v == fmt.Sprint(userIDClaim) {
-				return
-			}
+			err = errors.New("id_token is no valid")
```
babygoat: minor. "id_token is invalid"

babygoat:
Except for the wording in error message, other LGTM

PR #206, opened 2019-01-22
Fix activation url error

babygoat:
This patch fixes the incorrect url endpoint within member activation
mail.


babygoat:
@nickhsine

nickhsine:
LGTM

babygoat:
Clean rebased.

Issue #209, opened 2019-01-31
[Security] GET method on donation endpoints should use `user_id` in cookie rather than query param


nickhsine:
### Current Situation 
`/periodic-donations/:id`
`/donations/prime/:id`
`/donations/others/:id`
are using `user_id` query param.
For example, the client would send the GET method request like following
```
Headers:
  Cookie: id_token=<id_token>
  Authorization: Bearer <access_token>

Endpoint:
  `https://go-api.twreporter.org/v1/periodic-dontaions/10?user_id=1`
```

Even though we uses `middlewares/jwt.go` `ValidateAuthentication` and `ValidateAuthorization` to valid the request. However, we are only checking `donation.userID == c.Query('user_id')`, not checking `donation.userID == user_id in the access_token.payload`.
Therefore, hackers can send malicious requests to get others' donation data.

### Solution
Drop `user_id` query param.
Use `user_id in the access_token.payload` to validate the authorization.

babygoat:
Resolved in #211

PR #211, opened 2019-02-12
Auth improvement

babygoat:


babygoat:
@nickhsine

babygoat:
Address #209

tests/controller_donations_test.go
```diff
@@ -887,12 +881,6 @@ func TestGetAPeriodicDonationOfAUser(t *testing.T) {
 		assert.Equal(t, http.StatusUnauthorized, resp.Code)
 	})
 
-	t.Run("StatusCode=StatusForbidden", func(t *testing.T) {
-		path := fmt.Sprintf("/v1/periodic-donations/%d?user_id=%d", tokenRes.Data.ID, getUser(Globs.Defaults.Account).ID)
-		resp := serveHTTPWithCookies("GET", path, "", "application/json", authorization, cookie)
-		assert.Equal(t, http.StatusForbidden, resp.Code)
-	})
-
```
nickhsine: If we remove these two test cases, would we still test forbidden situations?

nickhsine:
LGTM except test cases removal.

tests/controller_donations_test.go
babygoat: Once we retrieve the user_id from query string, should we still need the forbidden case validation?

For the forbidden case, the jwt encryption key would be compromised so that it would be used to generate another user's token. I was wondering if this test case practical?

nickhsine: Does forbidden mean `someone don't have the right to get something`?
Even we retrieve the `user_id` from query string, we still need to check if `the user_id has the right to get the donation` or not, don't we?

Since we are going to return `status.Forbidden` in the function, I thought these test cases are still useful.

babygoat: Sorry there's a wrong statement in previous comment. "once we retrieve the user id from `authorization header`, ...."

I ignore the scenario that a valid user using a valid token to did the malicious query(e.g, query records from other users). I will update the tests back.

babygoat:
Updated.

nickhsine:
LGTM

PR #212, opened 2019-02-18
Bug fix. Escape query param in `ActivateLink`

nickhsine:
This patch fixes `ActivateLink` with unescaped characters.

@babygoat

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -334,7 +334,13 @@ func (mc *MembershipController) SignInV2(c *gin.Context) (int, gin.H, error) {
 	err = postMailServiceEndpoint(activationReqBody{
 		Email: email,
 		ActivateLink: fmt.Sprintf("%s://%s:%s/v2/auth/activate?email=%s&token=%s&destination=%s",
-			globals.Conf.App.Protocol, activateHost, globals.Conf.App.Port, email, activeToken, signIn.Destination),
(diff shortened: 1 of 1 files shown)


controllers/account.go
```diff
@@ -334,7 +334,13 @@ func (mc *MembershipController) SignInV2(c *gin.Context) (int, gin.H, error) {
 	err = postMailServiceEndpoint(activationReqBody{
 		Email: email,
 		ActivateLink: fmt.Sprintf("%s://%s:%s/v2/auth/activate?email=%s&token=%s&destination=%s",
-			globals.Conf.App.Protocol, activateHost, globals.Conf.App.Port, email, activeToken, signIn.Destination),
+			globals.Conf.App.Protocol,
+			activateHost,
+			globals.Conf.App.Port,
+			url.QueryEscape(email),
+			url.QueryEscape(activeToken),
+			url.QueryEscape(signIn.Destination),
```
babygoat: If we do the query escape here, should we unescape the string within `ActivateV2()` in `account.go`?

nickhsine: `gin.Context.Query` will run `QueryUnescape` by default.
Hence, we don't have to unescape query param by ourselves.

babygoat: Sure.

babygoat:
LGTM

PR #214, opened 2019-02-20
Only development environment can return draft posts or topics

nickhsine:
### Problem description
Behavior: 
We can fetch draft pots/topics from staging-go-api.twreporter.org.

Reason:
In the latest configs, the environment could be `development`, `staging` or `production`.
However, the codes only prevent `production` environment from fetching draft topics/posts.
Hence, in the `staging` environment, we could fetch the drat ones.

### Change log
Only allow development environment to return draft posts/topics.


babygoat:
Add reasons for this changes in commit message?

nickhsine:
Added

babygoat:
LGTM

PR #216, opened 2019-03-07
Update donation table schema

babygoat:
This patch enlarges the card_info_level size from varchar(10) to
varchar(255). Also, reorder the schema fields so that the order could
reflect the update history.


nickhsine:
LGTM

babygoat:
Clean rebase.

PR #217, opened 2019-03-14
Update go vendor package version

babygoat:
This patch updates the go vendor package to the lastest commit. Majorly
, the changes are to adopt the fix in gorm@1.9.2.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ jobs:
   build:
     docker:
       # specify the version you desire here
-      - image: thereporter/circleci-go-api:latest
+      - image: thereporter/circleci-go-api:1.10-alpine3.8
       - image: circleci/mongo:3.2
       - image: circleci/mysql:5.7
         environment:
--- a/glide.lock
+++ b/glide.lock
@@ -1,19 +1,19 @@
 hash: 2f8b4f0d6800da4322d085afb7e417a82270b936b830ce7026c80bc75382de98
-updated: 2018-09-26T19:24:29.041010934+08:00
+updated: 2019-03-14T19:17:03.775958453+08:00
 imports:
 - name: cloud.google.com/go
-  version: 97efc2c9ffd9fe8ef47f7f3203dc60bbca547374
+  version: f52f9bc132541d2aa914f42100c36d10b1ef7e0c
   subpackages:
   - compute/metadata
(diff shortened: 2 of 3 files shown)


babygoat:
@nickhsine 

There are two problems for the changes. 
1. gorm@1.9.2 requires sync.Map (concurrent-safe data structure) in go 1.9.0^. We will be required to upgrade the CI images for the go versions.
2.  glide lacks of the specific package upgrade until now.  I suggest moving forward to go module in replace of glide.

I would like to know the comments from you.

nickhsine:
@babygoat The circle build is failing. Please help to check the problem.

nickhsine:
Could we update `glide.yaml` instead of updating `glide.lock` which is a automatically generated by `glide`?

babygoat:
Please see my comment.
The error results from bullet 1. I'd like to know if you have any concerns of that.

babygoat:
> Could we update `glide.yaml` instead of updating `glide.lock` which is a automatically generated by `glide`?

Please see my bullet 2.
Sure. However, there's a lot of versions need to fix in glide.yaml if we'd like to adopt changes. 
If we'd like to upgrade the go environment now, my suggestion was also introduce the `go mod` instead of `glide` in 1.11 which might have better package management (e.g, single package upgrade commands or options.

nickhsine:
> > Could we update `glide.yaml` instead of updating `glide.lock` which is a automatically generated by `glide`?
> 
> Please see my bullet 2.
> Sure. However, there's a lot of versions need to fix in glide.yaml if we'd like to adopt changes.
> If we'd like to upgrade the go environment now, my suggestion was also introduce the `go mod` instead of `glide` in 1.11 which might have better package management (e.g, single package upgrade commands or options.

Sure, we could upgrade to  `go@1.11`.
However, it takes time. 
We might need to plan it.

In order to solve this dependency update, how about updating `glide.yaml` first? 
`gorm@^1.0.0` -> `gorm@^1.9.2` ?

For bullet 1, I have built a new [docker image circle-ci-go:1.10-alpine3.8](https://cloud.docker.com/u/thereporter/repository/docker/thereporter/circleci-go-api).

May we use that image to address this problem at this time?

babygoat:
Updated.

nickhsine:
LGTM

PR #240, opened 2019-04-09
Support limited payment times

babygoat:
This patch adds the max_paid_times entry to support the limited times
recurrence payment. Also, synchronize the document to the current
implementation.


babygoat:
@nickhsine

nickhsine:
LGTM

PR #241, opened 2019-04-11
Use forward-proxy to make tappay request

babygoat:
This patch adds the forward-proxy in staging/production environment to
perform tappay transaction.


nickhsine:
LGTM

PR #246, opened 2019-05-09
Fix missing proxy config

babygoat:
This patch fixes the missing proxy config loading from environment
variables.


nickhsine:
LGTM

PR #262, opened 2019-07-03
Fix incorrect schema during donation test

babygoat:
This patch fixes the incorrect schemas of pay_by_card_token_donations.
It results from the incorrect placement of `PayInfo` model which should
be specifc fields within pay_by_prime_donations model.


babygoat:
@nickhsine @taylrj

taylrj:
LGTM

babygoat:
clean rebased.

PR #281, opened 2019-07-30
Fix missing transaction time when transaction fail

babygoat:
This patch validate the transaction time field of the tappay response to
fix the invalid datetime format to database.


nickhsine:
LGTM

PR #292, opened 2019-09-26
Query endpoint to the tappay server

babygoat:
This patch adds the endpoint document to describe the behavior of
transaction query on the tappay server.


babygoat:
@taylrj  This is the reference of the transaction query endpoint

doc/donation.apib
```diff
@@ -0,0 +1,61 @@
...
++ filters
+    + `order_number`: `twreporter-153371414230837160610` (required)
+
+### TappayTradeRecords
++ record_status: 0 (number)
+
+### TappayServerResponse
++ status: 0 (required, number)
++ msg: "" (required)
++ trade_records (array[TappayTradeRecords])
```
nickhsine: Since `filters` is required and `filters.order_number` is required, there will be at most one record matched. In that condition,
1.  do we need `TappayServerRequest.records_per_page` ?
2. do we need `TappayServerResponse.trade_records`  to be an array?

babygoat: For resiliency, I would like to keep the endpoint as a structure relay for the time being. If we won't plan to add any other search ability for other use cases, this could be optimized.

nickhsine: Sure, we could this version first.
Modify it if needed in the future.

PR #294, opened 2019-10-04
Add refunded payment status

babygoat:
This patch adds the `refunded` payment status to track the payment
record which is refunded.


nickhsine:
LGTM

babygoat:
Clean rebase.

PR #300, opened 2019-10-18
Disable Docker Layer Caching

babygoat:
This patch disables the docker layer caching as the CircleCI pricing
plan change with the no free plan DLC support.


nickhsine:
LGTM

PR #316, opened 2019-11-11
Add linepay icon

babygoat:


babygoat:
line_pay_product_image_url validates the image size (exactly 84x84). Wait for the designer to provide the icon image.

PR #337, opened 2020-01-16
Config frontend host in runtime

babygoat:


babygoat:
Thanks for remind I will delete in later patch.

PR #363, opened 2020-03-04
Fix linepay notify format error

babygoat:


nickhsine:
LGTM.
Minor. Is this change a breaking change? Not just a bug fix? 
Would this change affect clients(support.twreporter.org)?

babygoat:
> LGTM.
> Minor. Is this change a breaking change? Not just a bug fix?
> Would this change affect clients(support.twreporter.org)?

Nope. But the request post body of route `/v1/donations/line-notify` is fixed. Does this just stand for the patch version update?

nickhsine:
> > LGTM.
> > Minor. Is this change a breaking change? Not just a bug fix?
> > Would this change affect clients(support.twreporter.org)?
> 
> Nope. But the request post body of route `/v1/donations/line-notify` is fixed. Does this just stand for the patch version update?
If this change doesn't affect clients,
I think we could bump patch version.

babygoat:
I got ur point. The public endpoint serves no change from API document but incorrect implementation. That should be a patch update.

babygoat:
Updated.

nickhsine:
LGTM

PR #368, opened 2020-03-10
Fix bookmark user identify

babygoat:


storage/bookmark.go
```diff
@@ -38,8 +38,7 @@ func (g *GormStorage) GetABookmarkByID(id string) (models.Bookmark, error) {
 // GetABookmarkOfAUser get a bookmark of a user
 func (g *GormStorage) GetABookmarkOfAUser(userID string, slug string, host string) (models.Bookmark, error) {
 	var bookmark models.Bookmark
-
-	err := g.db.Preload("Users", "id = ?", userID).Where("slug = ? and host = ?", slug, host).First(&bookmark).Error
+	err := g.db.Where("id in (?)", g.db.Table("users_bookmarks").Select("bookmark_id").Where("user_id = ?", userID).QueryExpr()).Where("slug = ? and host = ?", slug, host).First(&bookmark).Error
```
nickhsine: Minor. How about write plain SQL query?
Since we are not using gorm many-to-many feature, it would be clearer writing plain SQL query.

PR #374, opened 2020-05-14
Mongo read skew

babygoat:


utils/db.go
```diff
@@ -66,11 +66,10 @@ func InitMongoDB() (*mgo.Session, error) {
...
 
-	// Since we don't have writes here and don't care about the consistency between Mongo Master and Slave,
-	// we choose Eventual mode here.
-	// The Eventual mode is the fastest and most resource-friendly,
-	// but is also the one offering the least guarantees about ordering of the data read and written.
-	session.SetMode(mgo.Eventual, true)
+	// As our mongo cluster comprises cost-effective solution(Replica set arbiter),
+	// use Nearest read concern(https://docs.mongodb.com/manual/core/read-preference/#nearest)
+	// to distribute the read load acorss primary and secondary node evenly.
+	session.SetMode(mgo.Nearest, true)
```
nickhsine, line 72: In the document, https://godoc.org/gopkg.in/mgo.v2#Mode, it says
```
Nearest Mode = 6 // Read from one of the nearest members, irrespective of it being primary or secondary.

// Read preference modes are specific to mgo:
Eventual  Mode = 0 // Same as Nearest, but may change servers between reads.
```

Hence, I am not sure what this change is for.

babygoat: I have already seen this. However, under the hood, you will find SecondaryPreferred was configured within socket.go. Also, use global session for the mongo will increase the delay.

nickhsine:
LGTM 👍

PR #378, opened 2020-06-01
doc: update docs to add `receipt_header` field

taylrj:


doc/periodic-donation.apib
```diff
@@ -26,6 +26,7 @@ The Periodic Donation resource has the following attributes;
 - cardholder.national_id
 - cardholder.phone_number
 - cardholder.zip_code
+- cardholder.receipt_header
```
babygoat, line 29: We should promote the `receipt_header` to the top-level fields.

babygoat:
Can we have the donors leave blank to their receipt header?

taylrj:
Yes, receipt header is only required when donors need to receive receipt.

doc/periodic-donation.apib
taylrj: Updated.

doc/periodic-donation.apib
```diff
@@ -138,13 +140,14 @@ The states *id* and *order_number* are assigned by the TWReporter Go API at the
                         "name": "name(string) is optional",
                         "zip_code": "zip_code(string) is optional",
                         "phone_number": "phone_number(string) is optional",
-                        "national_id": "national_id(string) is optional"
+                        "national_id": "national_id(string) is optional",
```
babygoat, line 143: Is trailing comma a valid json format?

taylrj: No, it isn't. Next commit fixes the format.

babygoat:
LGTM

PR #380, opened 2020-06-03
chore: donations table schema change

taylrj:
This patch updates schemas of table `pay_by_prime_donations`,
`periodic_donations`, `pay_by_card_token_donations`,
'pay_by_other_method_donations' by adding `receipt_header` column.
Add corresponding `000006_add_receipt_header.down.sql` and
`000006_add_receipt_header.up.sql` for migration.

### DIFF
--- a/migrations/000006_add_receipt_header.up.sql
+++ b/migrations/000006_add_receipt_header.up.sql
@@ -1,0 +1,4 @@
+ALTER TABLE `pay_by_prime_donations` ADD `receipt_header` varchar(128) DEFAULT NULL;
+ALTER TABLE `periodic_donations` ADD `receipt_header` varchar(128) DEFAULT NULL;
+ALTER TABLE `pay_by_card_token_donations` ADD `receipt_header` varchar(128) DEFAULT NULL;
+ALTER TABLE `pay_by_other_method_donations` ADD `receipt_header` varchar(128) DEFAULT NULL;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 - api/donation
   - Append utm tag to donation link
+  - Add `receipt_header` column
 
 - api/mail
   - Update footer of the email template
--- a/migrations/000006_add_receipt_header.down.sql
+++ b/migrations/000006_add_receipt_header.down.sql
@@ -1,0 +1,4 @@
+ALTER TABLE `pay_by_prime_donations` DROP `receipt_header`;
+ALTER TABLE `periodic_donations` DROP `receipt_header`;
+ALTER TABLE `pay_by_card_token_donations` DROP `receipt_header`;
+ALTER TABLE `pay_by_other_method_donations` DROP `receipt_header`;


migrations/000006_add_receipt_header.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE `pay_by_prime_donations` ADD `receipt_header` varchar(30) DEFAULT NULL;
+ALTER TABLE `periodic_donations` ADD `receipt_header` varchar(30) DEFAULT NULL;
+ALTER TABLE `pay_by_card_token_donations` ADD `receipt_header` varchar(30) DEFAULT NULL;
+ALTER TABLE `pay_by_other_method_donations` ADD `receipt_header` varchar(30) DEFAULT NULL;
```
nickhsine, line 4: Just have a concern.
Is `varchar(30)` enough for `收據抬頭`?
I guess `receipt_header  varchar(30)` comes from `cardholder_name`.

In general, `cardholder_name` is hardly more than 30 characters.
However, `收據抬頭` might not be the same, especially if someone inputs english characters in `收據抬頭`.

I suggest that we can double the figure, like `varchar(64)`.
In addition, maybe mention the character number limit in the frontend, and provide instructions if user's input of `收據抬頭` is more than the constraint.

How do you think?

nickhsine: Provide an example,
our company name in english  `The Reporter Cultural Foundation` is more than 30 characters.

I guess we can increase it to `varchar(128)` for safety.

taylrj: Updated.

I think frontend instruction is also required as mentioned, I'll add it to frontend changes.

nickhsine:
LGTM 👍

migrations/000006_add_receipt_header.up.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE `pay_by_prime_donations` ADD `receipt_header` varchar(128) DEFAULT NULL;
+ALTER TABLE `periodic_donations` ADD `receipt_header` varchar(128) DEFAULT NULL;
+ALTER TABLE `pay_by_card_token_donations` ADD `receipt_header` varchar(128) DEFAULT NULL;
```
babygoat, line 3: Why do we have recipt_header for each pay_by_card_token_donations?

taylrj: For users who change the `receipt_header`, we can keep logs contain the unchanged ones in a year.

babygoat: The usage of `receipt_header` is for accounting agent to provide receipt according to desired frequency. Why do we keep the change revision? What is the use case of dated `receipt_header`?

taylrj: I believe this is an issue related to business logic and customer service.

Currently, we issue only one receipt and send it to donor once a year.
If we keep the change revisions, we can handle use case which needs receipt issued by month but also sent once a year.

However,  I have not seen  such use case so far. If there's no need to handle this case according to current business logic, I will create another patch to remove this column in table `pay_by_card_token_donations`.

babygoat: Yup, that was why I wanted to make sure that if it was designated feature from the accounting agent.
Besides, I raised the question here was to determine whether I should put the field within the `pay_by_card_token_donations`.

taylrj: I create a [patch](https://github.com/twreporter/go-api/pull/407) for it.

As far as I know, It was not a designated feature.
But before we make the call, I'd like to confirm with our accounting agent if it is a use case we need to handle.

taylrj: After having confirmed with accounting agent, we don't need to keep the change revisions. The edge cases, which might need multiple receipt headers should be handled individually.

PR #382, opened 2020-06-09
Features/news v2

babygoat:


babygoat:
The rewrite tasks still require sometime to work on.

The major components including
- [x] GetAPost
- [x] GetPosts (lack of where search)
- [x] GetATopic
- [x] GetTopics(lack of where search)
[] GetIndexPage
[] GetIndexPageCategories

Current setup is listed below.

Setup 
- macOS: 10.14.6
- docker engine: 19.03.4 (client&server)
- mongo standalone container server v3.6

專題文章
https://www.twreporter.org/a/photo-hong-kong-2019-protest-of-extradition-bill

- meta:
  - v1:  ~40ms
  - v2: ~12ms
- full:
  - v1: ~1.2s
  - v2: ~0.8s

非專題文章
https://www.twreporter.org/a/kaohsiung-han-kuo-yu-recall

- meta:
   - v1: ~60ms
   - v2: ~15ms
- full:
   - v1: ~260ms
    -v2: ~360ms

專題頁landing page
https://www.twreporter.org/topics/hong-kong-extradition-law

- meta
  - v1:  ~15ms
  - v2:  ~3 ms
- full
  - v1: 1~1.1s
  - v2:  0.45 s

Some insight from the rewrite can be grabbed from the v2 draft implementation

1. Overall, the latency improve around 33% in full mode and 70~80% in meta mode with the aggregation rewrite instead of multiple querys.
2. The more the related documents within topic/related fields, the more improvement with the rewrite

babygoat:
More performance tests

index_page on 6/05
- v1: ~1.2 s
- v2 ~0.6 s

loadtesting on the 6/05 dataset
- v1
```
Requests      [total, rate, throughput]         1500, 50.03, 5.54
Duration      [total, attack, wait]             37.707s, 29.981s, 7.726s
Latencies     [min, mean, 50, 90, 95, 99, max]  92.944µs, 4.522s, 5.317s, 5.976s, 7.138s, 22.863s, 29.94s
Bytes In      [total, mean]                     371769, 247.85
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           13.93%
Status Codes  [code:count]                      0:328  200:209  500:963
```

- v2
```
Requests      [total, rate, throughput]         1500, 50.04, 12.76
Duration      [total, attack, wait]             40.69s, 29.977s, 10.712s
Latencies     [min, mean, 50, 90, 95, 99, max]  106.257µs, 3.779s, 1.053s, 11.161s, 13.702s, 17.225s, 19.54s
Bytes In      [total, mean]                     4170840, 2780.56
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           34.60%
Status Codes  [code:count]                      0:285  200:519  500:696
```

Note that the success rate improves from (14% -> 35%).  The major bottleneck of the local load testing is the limited file descriptors (i.e., socket, see `ulimit -n`)  which greatly affects the response time.

babygoat:
Listing tests

posts (~2600)

- v1: ~180ms
- v2: ~50ms (with createIndex on publishedDate)

topics
- v1: ~70ms
- v2: ~10ms

Note that the createIndex on the dataset will drastically improve the operations performance (match/sort). This should be done along with the cluster migration.

babygoat:
@nickhsine  you can start reviewing from commit  6bbdf17. Those pre-commits serve as the proof-of-concept of using the mongo-go-driver and performance testing but lack of full structure design.

controllers/news_v2.go
```diff
@@ -106,6 +106,50 @@ func (nc *newsV2Controller) GetTopics(c *gin.Context) {
...
+		if err != nil {
+			switch {
+			case errors.Is(err, context.DeadlineExceeded):
+				c.JSON(http.StatusGatewayTimeout, gin.H{"status": "error", "message": "Query upstream server timeout."})
+			default:
+				c.JSON(http.StatusInternalServerError, gin.H{"status": "error", "message": "Unexpected error."})
+			}
+			log.Errorf("%+v", err)
+		}
+	}()
```
nickhsine, line 143: extract this anonymous function?
It seems that we are going to use the same logic to handle timeout and internal server error.

storage/news_v2.go
```diff
@@ -117,12 +117,109 @@ func (m *mongoStorage) getMetaOfPosts(ctx context.Context, stages []bson.D) <-ch
...
+	case <-ctx.Done():
+		return nil, errors.WithStack(ctx.Err())
+	case result, ok := <-m.getFullTopics(ctx, stages):
+		switch {
+		case !ok:
+			return nil, errors.WithStack(ctx.Err())
+		case result.Error != nil:
+			return nil, errors.WithStack(result.Error)
+		}
+		topics = result.Content.([]news.Topic)
```
nickhsine, line 140: Do we need to handle type assertion panic here?

storage/news_v2.go
```diff
@@ -117,12 +117,109 @@ func (m *mongoStorage) getMetaOfPosts(ctx context.Context, stages []bson.D) <-ch
...
+
+		var topics []news.Topic
+		for cursor.Next(ctx) {
+			var topic news.Topic
+			err := cursor.Decode(&topic)
+			if err != nil {
+				result <- fetchResult{Error: errors.WithStack(err)}
+			}
+			topics = append(topics, topic)
+		}
```
nickhsine, line 167: Do we need to check `cursor.Err()` here?

storage/news_v2.go
```diff
@@ -117,12 +117,109 @@ func (m *mongoStorage) getMetaOfPosts(ctx context.Context, stages []bson.D) <-ch
...
+
+		var topics []news.MetaOfTopic
+		for cursor.Next(ctx) {
+			var topic news.MetaOfTopic
+			err := cursor.Decode(&topic)
+			if err != nil {
+				result <- fetchResult{Error: errors.WithStack(err)}
+			}
+			topics = append(topics, topic)
+		}
```
nickhsine, line 217: Do we need to check `cursor.Err()` here?

controllers/news_v2.go
nickhsine: Please ignore this comment.
The following commits do the extraction already.

storage/news_v2.go
babygoat: I believed the switch statements above catches all the error cases AFAIK.
`case !ok` catches the context timeout if the storage layer returned early.
`result.Error` catches the operation errors through the api.

Was there any case I miss?

storage/news_v2.go
babygoat: Good gotcha! I forgot the error handling here.

controllers/news_v2.go
```diff
@@ -171,7 +173,211 @@ func (nc *newsV2Controller) GetATopic(c *gin.Context) {
...
+
+func (nc *newsV2Controller) getIndexPageJobs() []job {
+	const statePublished = "published"
+
+	// v1 index page section jobs
+	jobs := []job{
+		{
+			Name:  news.LatestSection,
+			Type:  typePost,
+			Query: news.NewQuery(news.FilterState(statePublished), news.Limit(6)),
```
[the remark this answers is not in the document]
babygoat: This is actually a idiomatic way to use called `functional options` pattern as promoted by one of the golang creator, [Rob Pike](https://commandcenter.blogspot.com/2014/01/self-referential-functions-and-design.html), and [Dave Cheney](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis). The reason why this pattern is preferred to `method chaining` in golang is that other language like java or python can throw error in the middle of the chain. However, golang code tends to return error rather throw error in the middle. Since the chaining method requires the function to return receiver itself, there is no way to break in the middle of chain but to record error within the receiver object (e.g., gorm.db.Error in gorm).

In our case, the goal is to create an immutable object pointer(*Query). There is no such drawback from the aforementioned issue so the decision to use functional option is subjective and I believed this is more like an overloading constructor (`func NewQuery(options func(*Query)) *Query`). Chaining methods, instead, are setters which might be sort of contradict to our goal semantically.

Functional option do have some drawbacks (e.g., package api bloating/not easy to learn what to configure) as discussed in the [google-cloud-go issue](https://github.com/google/go-cloud/issues/908) . I'm not sure if the problems you referred to the package level api are within these issues scope. In the former issue, I cannot foresee the requirements for adding more filters or sort but the current number (7) seems OK. In the latter one,  I think I can define the type of func(*Query) (i.e, `type Option func(*Query)`) so that these functions will be grouped under Option in godoc. What's more, I can prefix the function with *With* to indicate that this is actually a option function. (e.g., `func WithLimit(limit int) Option`.

What do u think?

storage/news_v2.go
nickhsine: Oops, I didn't comment clearly.

Do we need to add the second returned value `ok` here?

Such as
```
topics, ok = result.Content.([]news.Topic)
```

If we don't add `ok`, do we need to handle the type assertion panic?

babygoat: I omit the ok in purpose as the possible error cases are handled in the above switch cases.

As I proposed before, does there any case I miss?

controllers/news_v2.go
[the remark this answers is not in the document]
nickhsine: Use `type Option func(*Query)` to group these functions under Option in godoc sounds pretty good!
And adding `With` prefix is a good idea as well.

storage/news_v2.go
nickhsine: There is no cases we miss here.
Adding `ok` is a minor suggestion.

The reason I prefer to `ok` here is that 
we can make sure type assertion would not cause panic.
However, if we are sure there is no type assertion error here, we could skip `ok`.

storage/news_v2.go
```diff
@@ -0,0 +1,255 @@
...
+	// build lookup(join) stages according to required fields
+	stages = append(stages, news.BuildLookupStatements(news.LookupFullPost)...)
+
+	select {
+	case <-ctx.Done():
+		return nil, errors.WithStack(ctx.Err())
+	case result, ok := <-m.getFullPosts(ctx, stages):
+		if !ok {
+			return nil, errors.WithStack(ctx.Err())
+		}
```
nickhsine, line 42: miss `result.Error != nil` case?

storage/news_v2.go
```diff
@@ -0,0 +1,255 @@
...
+	// build lookup(join) stages according to required fields
+	stages = append(stages, news.BuildLookupStatements(news.LookupMetaOfPost)...)
+
+	select {
+	case <-ctx.Done():
+		return nil, errors.WithStack(ctx.Err())
+	case result, ok := <-m.getMetaOfPosts(ctx, stages):
+		if !ok {
+			return nil, errors.WithStack(ctx.Err())
+		}
```
nickhsine, line 89: miss `result.Error != nil` case?

storage/news_v2.go
```diff
@@ -0,0 +1,255 @@
...
+	stages := news.BuildQueryStatements(mq)
+	// build lookup(join) stages according to required fields
+	stages = append(stages, news.BuildLookupStatements(news.LookupFullPost)...)
+
+	select {
+	case <-ctx.Done():
+		return nil, errors.WithStack(ctx.Err())
+	case result, ok := <-m.getFullPosts(ctx, stages):
+		if !ok {
+			return nil, errors.WithStack(ctx.Err())
```
nickhsine, line 41: Question 🙋 :
Does `ok == false` mean that `result` channel is closed?

storage/news_v2.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+		posts = result.Content.([]news.Post)
+	}
+
+	return posts, nil
+}
+
+func (m *mongoStorage) getFullPosts(ctx context.Context, stages []bson.D) <-chan fetchResult {
+	result := make(chan fetchResult)
```
nickhsine, line 50: minor.
use buffered channel prevent goroutine leaks in case the channel is never read ?
```
result := make(chan fetchResult, 1)
```

controllers/news_v2.go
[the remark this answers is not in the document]
babygoat: done

storage/news_v2.go
babygoat: done

storage/news_v2.go
babygoat: done

storage/news_v2.go
babygoat: done

storage/news_v2.go
babygoat: done

storage/news_v2.go
babygoat: Yes, the above error handling section tries to deal with the cases which the assertion would be failed. If these were done properly, then I believed validation here can be saved.

storage/news_v2.go
babygoat: Yes, the ok served as indicator whether the channel was closed or not. This assertion is idempotent.

storage/news_v2.go
babygoat: The result channel will be closed by the anonymous function below from another goroutine. As the context was passed to that goroutine, it would either return or timeout and then close the channel properly. Also, we only transfer single result not multiple results, hence, we don't need to declare the buffer channel.

babygoat:
Updated.

storage/news_v2.go
nickhsine: So, if the channel is closed, there will be an error?

babygoat: This is to deal with the case if timeout occurs and the `m.getFullPosts` returns earlier than the `ctx.Done()`.  However, since the `m.getFullPosts` always return result, it seems to me that this can be saved as well.

controllers/news_v2.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+		return jobs
+	}(
+		news.HumanRightsAndSociety,
+		news.EnvironmentAndEducation,
+		news.PoliticsAndEconomy,
+		news.CultureAndArt,
+		news.International,
+		news.LivingAndMedicalCare,
+	)...)
```
nickhsine, line 295: Minor.
```
    for _, v := range []news.Category{
      news.HumanRightsAndSociety,
      news.EnvironmentAndEducation,
      news.PoliticsAndEconomy,
      news.CultureAndArt,
      news.International,
      news.LivingAndMedicalCare,
    } {
      jobs = append(jobs, job{
        v.Name,
        typePost,
        news.NewQuery(news.WithFilterCategoryIDs(v.ID), news.WithLimit(1)),
      })
    }
```
?

controllers/news_v2.go
```diff
@@ -0,0 +1,388 @@
...
+			if result.Error != nil {
+				err = result.Error
+				break
+			}
+			results[result.Name] = result.Content
+			if len(results) == len(jobs) {
+				break
+			}
+		}
+	}
```
nickhsine, line 228: Do we have simpler way to fetch sections and categories individually?
Currently we use many channels in our logics, which is a little bit difficult to understand.

babygoat: Can you elaborate more details?

These are general and idiomatic concurrent patterns including fan-out, multiplexing(nc.prepareJobs, nc.fetchJobs) and fan-in (nc.merge). Here are some reference for these [common pattern in golang](https://blog.golang.org/pipelines)

babygoat:
Updated. Sorry for accidentally force-pushed the commits. Change starts from 9c32bde

controllers/news_v2.go
babygoat: updated.

nickhsine:
LGTM 👍 🥇 👍 🥇

controllers/news_v2.go
```diff
@@ -0,0 +1,342 @@
...
+		defer close(resultStream)
+		wg.Wait()
+	}()
+	return resultStream
+}
+
+func (nc *newsV2Controller) helperCleanup(c *gin.Context, err error) {
+	if err != nil {
+		switch {
+		case errors.Is(err, context.DeadlineExceeded):
```
nickhsine, line 335: Is `Is` declared by `errors` package?

babygoat: Yes, you should have at least go version > 1.13.0. That's why I upgraded the go runtime of the image to the latest version.

controllers/news_v2.go
```diff
@@ -0,0 +1,342 @@
...
+			results[result.Name] = result.Content
+			if len(results) == len(jobs) {
+				break
+			}
+		}
+	}
+	if err != nil {
+		return
+	}
+	c.JSON(http.StatusOK, gin.H{"status": "success", "data": gin.H{"records": results}})
```
nickhsine, line 231: According to doc (https://github.com/twreporter/go-api/blob/master/doc/news/index_page.apib#L26), 
 
`"data": gin.H{"records": results}` ->  `"data": results` ?

internal/news/query.go
```diff
@@ -0,0 +1,198 @@
...
+	q = defaultQuery
+	// Parse filter
+	if len(c.QueryArray(queryCategoryID)) > 0 {
+		q.Filter.Categories = c.QueryArray(queryCategoryID)
+	}
+	if len(c.QueryArray(queryTagID)) > 0 {
+		q.Filter.Tags = c.QueryArray(queryTagID)
+	}
+	if len(c.QueryArray(queryPostID)) > 0 {
+		q.Filter.Tags = c.QueryArray(queryPostID)
```
nickhsine, line 131: `q.Filter.Tags` -> `q.Filter.IDs`  ?

babygoat:
### setup:

* Testing datasets
   * 8/4 index_page
   * https://www.twreporter.org/a/saturday-features-mountain-climbing-trial-or-homesickness
   * https://www.twreporter.org/a/foster-care-system-adoption-flow-and-rule
   * https://www.twreporter.org/a/mini-reporter-recall
   * https://www.twreporter.org/a/covid-19-greek-refugee-camps
   * https://www.twreporter.org/a/2019-ncov-epidemic
   * https://www.twreporter.org/a/saturday-features-philosophy-wormhole-postulated-author

* Testing load
  * Avg: 100 req/s
  * Duration: 30 sec

### v1: 

timeout: 
* index_page: 3s
* post:5s

```sh
Requests      [total, rate, throughput]         3000, 100.03, 1.97
Duration      [total, attack, wait]             47.321s, 29.99s, 17.331s
Latencies     [min, mean, 50, 90, 95, 99, max]  23.687ms, 5.313s, 81.853ms, 30.001s, 30.002s, 30.002s, 30.005s
Bytes In      [total, mean]                     535186, 178.40
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           3.10%
Status Codes  [code:count]                      0:502  200:93  500:8  520:86  525:2311
Error Set:
525 [1]
520 [2]
500 Internal Server Error
```

### v2:
* index_page: 5s
* post: 5s

```sh
Requests      [total, rate, throughput]         3000, 100.03, 36.19
Duration      [total, attack, wait]             35.509s, 29.99s, 5.519s
Latencies     [min, mean, 50, 90, 95, 99, max]  23.235ms, 4.151s, 5.154s, 6.643s, 7.211s, 7.97s, 8.295s
Bytes In      [total, mean]                     12022075, 4007.36
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           42.83%
Status Codes  [code:count]                      200:1285  500:249  504:1219  520:50  525:197
Error Set:
504 Gateway Timeout
525
520
500 Internal Server Error
```

### v2:
* index_page: 6s
* post: 6s

```sh
Requests      [total, rate, throughput]         3000, 100.04, 99.71
Duration      [total, attack, wait]             30.086s, 29.989s, 96.756ms
Latencies     [min, mean, 50, 90, 95, 99, max]  23.453ms, 344.784ms, 48.036ms, 1.201s, 1.794s, 3.844s, 5.085s
Bytes In      [total, mean]                     56573488, 18857.83
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:3000
Error Set:
```

### Disuccsion
* Successful rate:  ~3%(v1) -> ~40%(v2) -> ~100%(v2 with proper timeout tuning)

PR #394, opened 2020-09-03
api/news: fix video bson document decoder

babygoat:
This patch fixes the video bson document decoder which should decode
`filetype`, `size`, `url` fields from the nested `video` document

i.e,

{...
  "leading_video": {
    "_id": ObjectID(...),
    "title": title,
    "video": {
      "filetype": ...,
      "size": ...,
      "url": ...,
    }
  }
...
}

### DIFF
--- a/internal/news/model.go
+++ b/internal/news/model.go
@@ -129,10 +129,14 @@ type Tag struct {
 	Name string             `bson:"name" json:"name"`
 }
 
+type VideoMeta struct {
+	Filetype string `bson:"filetype" json:"filetype"`
+	Size     uint   `bson:"size" json:"size"`
+	URL      string `bson:"url" json:"url"`
+}
+
 type Video struct {
-	ID       primitive.ObjectID `json:"id" bson:"_id"`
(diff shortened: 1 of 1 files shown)


internal/news/model.go
```diff
@@ -129,10 +129,14 @@ type Tag struct {
...
+	URL      string `bson:"url" json:"url"`
+}
+
 type Video struct {
-	ID       primitive.ObjectID `json:"id" bson:"_id"`
-	Title    string             `json:"title"`
-	Filetype string             `json:"filetype"`
-	Size     uint               `json:"size"`
-	URL      string             `json:"url"`
+	VideoMeta `bson:"video"`
```
nickhsine, line 139: `bson: "video" json:"video"` ?

babygoat: we don't plan to nested the `filetype`, `size`, `url` into `video`, are we?  Embedded struct does not unmarshal the root by default.

nickhsine: Yes, you are right, we don't nest it in the latest design.

PR #407, opened 2020-09-26
chore: `pay_by_card_token_donations` table schema change

taylrj:
This patch updates schema of table `pay_by_card_token_donations`
by removing `receipt_header` column.
Add corresponding
`000007_delete-receipt_header-from-pay_by_card_token_donations.down.sql`
and `000007_delete-receipt_header-from-pay_by_card_token_donations.up.sql`
for migration.


babygoat:
`Alter table Drop column` will result in table write lock and thus fail the recurrence donation during the processing time. I suggest we have the dummy column as is.

PR #408, opened 2020-10-14
fix: update membershio_user.sql to catch up the latest migrations

nickhsine:
### Reason to Change
So far, `twreporter/environment-setup`  builds local mysql database with `membership_user.sql` schema.
However, `membership_user.sql` is out of date.
Those new migrations are not applied to it.
This PR is to update the new schema.

### DIFF
--- a/membership_user.sql
+++ b/membership_user.sql
@@ -226,7 +226,7 @@ CREATE TABLE `pay_by_prime_donations` (
   `transaction_time` timestamp NULL DEFAULT NULL,
   `bank_transaction_start_time` timestamp NULL DEFAULT NULL,
   `bank_transaction_end_time` timestamp NULL DEFAULT NULL,
-  `bank_result_code` varchar(50) NULL DEFAULT NULL,
+  `bank_result_code` varchar(128) NULL DEFAULT NULL,
   `bank_result_msg` varchar(50) NULL DEFAULT NULL,
   `cardholder_email` varchar(100) NOT NULL,
   `cardholder_phone_number` varchar(20) DEFAULT NULL,
@@ -247,7 +247,8 @@ CREATE TABLE `pay_by_prime_donations` (
   `notes` varchar(100) DEFAULT NULL,
   `is_anonymous` tinyint(1) DEFAULT 0,
   `linepay_method` enum('CREDIT_CARD', 'BALANCE', 'POINT') DEFAULT NULL,
-  `linepay_point` int DEFAULT NULL, 
+  `linepay_point` int DEFAULT NULL,
+  `receipt_header` varchar(128) DEFAULT NULL,
 
   PRIMARY KEY (`id`),
   KEY `idx_pay_by_prime_donations_status` (`status`),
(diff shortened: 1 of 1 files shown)


taylrj:
LGTM

babygoat:
I suggest not proceeding with the approach. This will always incur similar schema out-of-sync issue. Instead, try to use the migrations file on master.

nickhsine:
> I suggest not proceeding with the approach. This will always incur similar schema out-of-sync issue. Instead, try to use the migrations file on master.

Using migration files is the better solution.
However, we need to update `twreporter/environment-setup`, and I haven't come up with a good approach to do so.

This PR is a quick fix to have the latest schema in local mysql database.

babygoat:
Here are the draft steps for doing the mysql build with the required migrations
1. In dockerfile, sparse checkout go-api/pangolin repository to obtain the migrations directories in the base image
2. In the target image, copy the directories from the base image
3. Download the [migrate binary](https://github.com/golang-migrate/migrate/releases/tag/v4.13.0) with executable permission
4. Perform the migrations

We always run out of time to do things in more robust way.

If you still require the quick-fix, just proceed.

nickhsine:
> Here are the draft steps for doing the mysql build with the required migrations
> 
> 1. In dockerfile, sparse checkout go-api/pangolin repository to obtain the migrations directories in the base image
> 2. In the target image, copy the directories from the base image
> 3. Download the [migrate binary](https://github.com/golang-migrate/migrate/releases/tag/v4.13.0) with executable permission
> 4. Perform the migrations
> 
> We always run out of time to do things in more robust way.
> 
> If you still require the quick-fix, just proceed.

Thanks for advice.
Due to not familiar with `migrate`, I have a question about this approach.
How to perform the migrations if mysql database is not ready for connection?

nickhsine:
Here is my draft steps, but I need time to implement it.
Step 1.-3. is the same as yours.
4. Add migration scripts under `/docker-entrypoint-initdb.d/` of the container.

babygoat:
> > Here are the draft steps for doing the mysql build with the required migrations
> > 
> > 1. In dockerfile, sparse checkout go-api/pangolin repository to obtain the migrations directories in the base image
> > 2. In the target image, copy the directories from the base image
> > 3. Download the [migrate binary](https://github.com/golang-migrate/migrate/releases/tag/v4.13.0) with executable permission
> > 4. Perform the migrations
> > 
> > We always run out of time to do things in more robust way.
> > If you still require the quick-fix, just proceed.
> 
> Thanks for advice.
> Due to not familiar with `migrate`, I have a question about this approach.
> How to perform the migrations if mysql database is not ready for connection?

You can use either [wait_for_it.sh](https://github.com/vishnubob/wait-for-it) or use [dockerize](https://github.com/jwilder/dockerize) in entrypoint.sh like [this](sh -c "dockerize -wait tcp://$GOAPI_DB_MYSQL_ADDRESS:$GOAPI_DB_MYSQL_PORT -timeout 30s")

PR #410, opened 2020-10-14
Add customer service notes

babygoat:
Address [[story-24][task-1] add `customer_service_notes` field on periodic_donations/prime_donations](https://kanbanflow.com/t/p24MkSz1)


babygoat:
[story-24] changes status to invalid after discussion with marketing.

PR #418, opened 2021-01-27
api/donations: return error if patch record failed

babygoat:
This patch returns the status 500 (Internal Server Error) if the
donation record patch failed.


taylrj (approved):
LGTM

PR #420, opened 2021-03-03
api/news: get authors from algolia in primary

babygoat:
This patch adds the primary data source of authors(algolia) and
fallbacks to the mongo query if the index query results in error.


babygoat:
Updated. Ready for review. Besides, the index data are uploaded to `developer-for-testing` application which can be utilized to integration testing.

controllers/news_v2.go
```diff
@@ -358,18 +359,22 @@ func (nc *newsV2Controller) GetAuthors(c *gin.Context) {
 	q := news.ParseAuthorListQuery(c)
 
 	//TODO(babygoat): fetch from algolia first
```
taylrj: Remove TODO comment ?

internal/news/author.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/algolia/algoliasearch-client-go/v3/algolia/opt"
+	"github.com/pkg/errors"
+)
+
+func GetAuthorWithIndex(index AlgoliaSearcher, q *Query) ([]Author, int, error) {
+	var authors []Author
+	res, err := index.Search(q.Filter.Name, opt.Offset(q.Offset), opt.Length(q.Limit))
+	if err != nil {
+		// fallback
+		return nil, -1, errors.WithStack(err)
```
taylrj, line 16: Q: Why does it returns `-1` instead of `0` as the `nbHits` return value ?

storage/news_v2.go
```diff
@@ -316,33 +316,52 @@ func (m *mongoStorage) getAuthors(ctx context.Context, stages []bson.D) <-chan f
...
+	case <-ctx.Done():
+		return 0, errors.WithStack(ctx.Err())
+	case res, ok := <-result:
+		switch {
+		case !ok:
+			return 0, errors.WithStack(ctx.Err())
+		case res.Error != nil:
+			return 0, res.Error
+		}
+		count = res.Content.(int64)
```
taylrj, line 363: Do we need to check whether the assertion is ok here ?

e.g.

```go
  if count, ok := res.Content.(int64); ok {
     return count, nil
  }
```

babygoat:
updated.

storage/news_v2.go
babygoat: We ensure the `res.Content` is always int64 by single sender(Ln350).

internal/news/author.go
babygoat: updated.

babygoat: BTW, `-1` is used to mark the value is valid whenever the `err != nil`.

babygoat:
Merge first for demo.

taylrj:
LGTM

PR #433, opened 2021-07-15
api/donation: record api update

babygoat:
Address [twreporter-105](https://twreporter-org.atlassian.net/browse/TWREPORTER-105)

### DIFF
--- a/controllers/donation.go
+++ b/controllers/donation.go
@@ -217,8 +217,16 @@ type (
 		ReceiptHeader null.String       `json:"receipt_header"`
 	}
 
+	queryFilterTime struct {
+		StartTime null.Int `json:"start_time"`
+		EndTime   null.Int `json:"end_time"`
+	}
+
 	queryFilter struct {
-		OrderNumber string `json:"order_number" binding:"required"`
+		OrderNumber       string           `json:"order_number" binding:"required"`
+		BankTransactionID null.String      `json:"bank_transaction_id"`
+		RecTradeID        null.String      `json:"rec_trade_id"`
+		Time              *queryFilterTime `json:"time"`
 	}
 
 	queryReq struct {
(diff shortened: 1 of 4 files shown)


controllers/donation.go
```diff
@@ -941,6 +949,13 @@ func (mc *MembershipController) QueryTappayServer(c *gin.Context) (int, gin.H, e
 		}}, nil
 	}
 
+	// If the required fields `bank_transaction_id` or `rec_trade_id` are not specified, append the time filter of one month range from now
+	if reqBody.Filters.RecTradeID.IsZero() && reqBody.Filters.RecTradeID.IsZero() {
```
Aylie-Chou, line 953: one RecTradeId should be BankTransactionID

controllers/donation.go
```diff
@@ -941,6 +949,13 @@ func (mc *MembershipController) QueryTappayServer(c *gin.Context) (int, gin.H, e
 		}}, nil
 	}
 
+	// If the required fields `bank_transaction_id` or `rec_trade_id` are not specified, append the time filter of one month range from now
```
Aylie-Chou, line 952: Should we also check the validation of StartTime & EndTime ?

tests/controller_donations_test.go
```diff
@@ -1690,6 +1741,37 @@ func TestQueryTappayServer(t *testing.T) {
...
+	var rb recordRequestBody
+	err = json.Unmarshal(b, &rb)
+	if err != nil {
+		return false
+	}
+
+	// If the request filter does not provide either `rec_trade_id` or `bank_transaction_id` filter,
+	// the time filter should be set.
+	if rb.Filters.RecTradeID.IsZero() && rb.Filters.BankTransactionID.IsZero() {
+		if rb.Filters.Time.EndTime.IsZero() || rb.Filters.Time.StartTime.IsZero() {
```
Aylie-Chou, line 1768: From tappay doc, time filter should be within 90 days.
Should validation function mark it invalid if time filter is provide but exceed the 90 days limit ?

controllers/donation.go
babygoat: This is a bit tricky here. TapPay will not return any error even if the range is incorrect (i.e exceed 90 days). It will only return the result within the range. What's more, since the `time` filter is not exported (i.e. will be overwritten by go-api), I just omitted the verification here.

tests/controller_donations_test.go
babygoat: See above

controllers/donation.go
babygoat: Updated.

Aylie-Chou (approved):
LGTM

taylrj:
LGTM

PR #435, opened 2021-07-20
api/auth: log error if the session cannot be saved

babygoat:
This patch adds the error log if the session used for the oauth
verification cannot be preserved. With the error log, the error can be
reported by the google cloud error reporting service so that it can be
handled as soon as possible.


babygoat:
Address [twreporter-86](https://twreporter-org.atlassian.net/browse/TWREPORTER-86)

Aylie-Chou (approved):
LGTM

PR #440, opened 2021-07-27
api/donation: fix time filter unit

babygoat:
This patch correct the time filter(start_time/end_time) unit which
should be millisecond rather than second.
ref: https://docs.tappaysdk.com/tutorial/zh/reference.html#time

Also, increase the time filter range to 90 days according to the document
ref: https://docs.tappaysdk.com/tutorial/zh/reference.html#time-filter

### DIFF
--- a/controllers/donation.go
+++ b/controllers/donation.go
@@ -950,12 +950,12 @@ func (mc *MembershipController) QueryTappayServer(c *gin.Context) (int, gin.H, e
 		}}, nil
 	}
 
-	// If the required fields `bank_transaction_id` or `rec_trade_id` are not specified, append the time filter of one month range from now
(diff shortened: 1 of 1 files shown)


controllers/donation.go
```diff
@@ -949,12 +949,12 @@ func (mc *MembershipController) QueryTappayServer(c *gin.Context) (int, gin.H, e
 		}}, nil
 	}
 
-	// If the required fields `bank_transaction_id` or `rec_trade_id` are not specified, append the time filter of one month range from now
+	// If the required fields `bank_transaction_id` or `rec_trade_id` are not specified, append the time filter of the three month range from now
```
Aylie-Chou, line 952: Three month can be specific to 90 days

babygoat: Updated.

Aylie-Chou (approved):
LGTM

PR #451, opened 2021-09-28
api/donation: fix patch of Line POINT only transaction 

babygoat:
Address [TWREPORTER-156](https://twreporter-org.atlassian.net/browse/TWREPORTER-156)

### DIFF
--- a/controllers/donation.go
+++ b/controllers/donation.go
@@ -870,17 +870,6 @@ func (mc *MembershipController) PatchLinePayOfAUser(c *gin.Context) (int, gin.H,
 		}
 	}
 
-	if linePayMethodCreditCard == callbackPayload.PayInfo.Method.String {
-		// Validate Line Pay Masked Credit Card Number format
-		// sample: ************1234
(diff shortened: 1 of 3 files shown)


controllers/donation.go
```diff
@@ -869,17 +870,6 @@ func (mc *MembershipController) PatchLinePayOfAUser(c *gin.Context) (int, gin.H,
 		}
 	}
 
-	if linePayMethodCreditCard == callbackPayload.PayInfo.Method.String {
```
Aylie-Chou, line 872: Should we still check credit card number validation when user use credit card in line pay ?

babygoat: The problem lies in that the method `CREDIT_CARD` can represent three combinations (credit_card only, credit_card+line point(partial), line point only. In the latter case(line point only), the credit card number will not be returned (empty string only).

babygoat: More detail in Jira.

Aylie-Chou (approved):
LGTM

taylrj:
LGTM

PR #459, opened 2021-12-13
api/account: redirect to the error redirection URL when error occurs

taylrj:
Address [TWREPORTER-199](https://twreporter-org.atlassian.net/browse/TWREPORTER-199)

This patch redirects users to the customized error redirection URL
when error occurs on redirect to destination.

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -18,22 +18,17 @@ import (
 	"github.com/twreporter/go-api/utils"
 )
 
-const (
-	// TODO: Update later after UX designed
-	authErrorPage = "https://www.twreporter.org/"
+const idTokenExpiration = 60 * 60 * 24 * 30 * 6
 
-	defaultRedirectPage = "https://www.twreporter.org/"
-
-	idTokenExpiration = 60 * 60 * 24 * 30 * 6
-)
-
+var defaultRedirectPage = "https://www.twreporter.org/"
 var defaultPath = "/"
 
 func (mc *MembershipController) AuthByEmail(c *gin.Context, sendMailRoutePath string) (int, gin.H, error) {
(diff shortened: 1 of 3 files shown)


controllers/account.go
```diff
@@ -172,11 +170,17 @@ func (mc *MembershipController) ActivateV2(c *gin.Context) {
 	email := c.Query("email")
 	token := c.Query("token")
 	destination := c.Query("destination")
+	errorRedirection := c.Query("error_redirection")
```
Aylie-Chou, line 173: Nameing question: why does /active accept errorRedirect as "error_redirection" while /authenticate accept it as "redirectOnError" ?

taylrj: It didn't have special idea in the naming decision.
I distinguish two of them because the errorRedirection get its value via query string, the usage seems not the same.

What do you say? Do you prefer to unify the name between them ?

Aylie-Chou: In my opinion, "error_redirection" seems to be better than "redirectOnError" due to its same part of speech with "destination".
If you think "redirectOnError" is more suitable for /authenticate, we can keep current version, otherwise unifing them as error_redirection / errorRedirection looks clear for me.

taylrj: > "error_redirection" seems to be better than "redirectOnError" due to its same part of speech with "destination".

That's a good reason, updated.

To address this issue, I updated https://github.com/twreporter/membership-fe/pull/506/commits/3ac1d99af8a9b5dc1328e42e0f7d351dbfa6d0f2 as well since we change the property name of request body. Please have a see.

Aylie-Chou:
LGTM

taylrj:
Clean rebased.

PR #473, opened 2022-03-14
feat: publish upsert donation info to cloud pub/sub

Aylie-Chou:
- use [cloudpub package](https://github.com/twreporter/go-mod-lib/blob/master/pkg/cloudpub/publish.go) in go-mod-lib
- configs/config.go: add publish related conf
- main.go: initialize cloudpub service


configs/config.go
```diff
@@ -76,19 +76,23 @@ news:
     topic_page_timeout: 5s
     index_page_timeout: 5s
     author_page_timeout: 5s
+neticrm:
+    project_id: "" # gcp project id
+    pub_topic: "" # pub/sub topic
```
taylrj, line 81: Question: does the pub_topic be `sync-to-niticrm` ?

image

Aylie-Chou: `sync-to-niticrm` topic is created for testing (and also a typo 😂 )
In production, the topic name would be `sync-neticrm` based on [design document](https://www.notion.so/twreporter/d36c052646c24c6db66b187c67f976f5).

main.go
```diff
@@ -72,6 +73,14 @@ func main() {
 	defer func() {
 		client.Disconnect(ctx)
 	}()
+
+	// init cloudpub client
+	pubConfig := &cloudpub.Config{
+		ProjectID: globals.Conf.Neticrm.ProjectID,
+		Topic: globals.Conf.Neticrm.Topic,
+	}
+	cloudpub.NewPublisher(context.TODO(), pubConfig)
```
taylrj: Question: If it is the root of context, why doesn't it send `context.Background()` into `NewPublisher` ? Just wondering the differences 🤔

taylrj (approved):
LGTM

PR #477, opened 2022-03-21
fix: add pubsub credentials

Aylie-Chou:
This patch add pubsub credentials to go-api image
- var `PUBSUB_CREDENTIALS` is defined in circleci env


taylrj (commented):
LGTM

PR #483, opened 2022-04-13
refactor: update age range in model_user, donation models

taylrj:
Closes [TWREPORTER-282](https://twreporter-org.atlassian.net/browse/TWREPORTER-282).


taylrj:
Would like to discuss with @Aylie-Chou about the indentation issue in this PR.

Aylie-Chou:
As the prior discussion, I would prefer not to migrate db field name in this case.
We could just modify the display name in front end to solve this issue.

PR #484, opened 2022-04-27
fix: fix periodic donation do not store pay_method field

Aylie-Chou:
This patch add gorm config to store pay_method field.

### DIFF
--- a/models/donation.go
+++ b/models/donation.go
@@ -41,7 +41,7 @@ type Cardholder struct {
 	Nickname           null.String `gorm:"column:cardholder_nickname;type:varchar(50)" json:"nickname"`
 	Title              null.String `gorm:"column:cardholder_title;type:varchar(30)" json:"title"`
 	LegalName          null.String `gorm:"column:cardholder_legal_name;type:varchar(50)" json:"legal_name"`
-	Gender             null.String `gorm:"column:cardholder_gender;type:varchar(2)" json:"gender"`     // e.g., "M", "F", "X, "U"
+	Gender             null.String `gorm:"column:cardholder_gender;type:varchar(2)" json:"gender"` // e.g., "M", "F", "X, "U"
 	AgeRange           null.String `gorm:"column:cardholder_age_range;type:ENUM('less_than_18', '18_to_24', '25_to_34', '35_to_44', '45_to_54', '55_to_64', 'above_65')" json:"age_range"`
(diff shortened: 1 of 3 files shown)


models/donation.go
```diff
@@ -143,25 +143,25 @@ type PeriodicDonation struct {
...
+	Notes            string     `gorm:"type:varchar(100)" json:"notes"`
+	OrderNumber      string     `gorm:"type:varchar(50);not null" json:"order_number"`
+	SendReceipt      string     `gorm:"type:ENUM('yearly', 'no', 'no_receipt', 'digital_receipt_by_year', 'paperback_receipt_by_year');default:'no_receipt'" json:"send_receipt"`
+	Status           string     `gorm:"type:ENUM('to_pay','paying','paid','fail','stopped','invalid');not null" json:"status"`
+	ToFeedback       null.Bool  `gorm:"type:tinyint(1);default:1" json:"to_feedback"`
+	UpdatedAt        time.Time  `json:"updated_at"`
+	UserID           uint       `gorm:"type:int(10) unsigned;not null" json:"user_id"`
+	IsAnonymous      null.Bool  `gorm:"type:tinyint(1);default:0" json:"is_anonymous"`
+	AutoTaxDeduction null.Bool  `gorm:"type:tinyint(1)" json:"auto_tax_deduction"`
+	PayMethod        string     `gorm:"type:ENUM('credit_card','line','apple','google','samsung')" json:"pay_method"`
```
taylrj, line 166: Question: Do we need to create a database index for `PayMethod` such as `PayMethod` in `PayByPrimeDonation` ?

Aylie-Chou: For now, periodic donation has only one pay method, thus index on pay method has no benefit.
If we want to add pay method in the future, there're more than index to do:
- controller/donation should support mutiple pay method (currently we use credit card in all place)
- pay method index
  - model/donation gorm config add index
  - migration file add index
  - migration file update pay method column (for data before this patch)
- add pay method test case

In this patch, we only deal with storing pay method into database, and I think futhur work could be done when we like to add other pay method.
What do say about it ?

taylrj: It's clear, thanks for explaining that to me.

taylrj (approved):
LGTM

howar31 (approved):
LGTM

PR #494, opened 2022-11-23
feat: Add category/subcategory related features

howar31:
Completed APIs:
- Add new API: GET /v2/tags/{latest_order}{offset}{limit}
- Change API: GET /v2/index_page

### DIFF
--- a/controllers/news_v2.go
+++ b/controllers/news_v2.go
@@ -20,6 +20,8 @@ type newsV2Storage interface {
 	GetMetaOfTopics(context.Context, *news.Query) ([]news.MetaOfTopic, error)
 	GetAuthors(context.Context, *news.Query) ([]news.Author, error)
 
+	GetTags(context.Context, *news.Query) ([]news.Tag, error)
+
 	GetPostCount(context.Context, *news.Query) (int64, error)
 	GetTopicCount(context.Context, *news.Query) (int64, error)
 	GetAuthorCount(context.Context, *news.Query) (int64, error)
@@ -56,6 +58,14 @@ func (nc *newsV2Controller) GetPosts(c *gin.Context) {
 
 	total, err := nc.Storage.GetPostCount(ctx, q)
 	if err != nil {
+		return
+	}
+
+	if q.Filter.SubcategoryID != "" && len(posts) == 0 {
+		c.JSON(http.StatusBadRequest, gin.H{
+			"status":  "error",
+			"message": "category & subcategory is not consistent",
+		})
 		return
 	}
 
@@ -105,6 +115,37 @@ func (nc *newsV2Controller) GetAPost(c *gin.Context) {
 	}
 
 	c.JSON(http.StatusOK, gin.H{"status": "success", "data": post})
+}
+
+func (nc *newsV2Controller) GetTags(c *gin.Context) {
+	var err error
+
+	ctx, cancel := context.WithTimeout(c, globals.Conf.News.PostPageTimeout)
+	defer cancel()
+
+	defer func() {
+		if err != nil {
+			nc.helperCleanup(c, err)
+		}
+	}()
+
+	q := news.ParseTagListQuery(c)
+
+	tags, err := nc.Storage.GetTags(ctx, q)
+
+	if err != nil {
+		return
+	}
+
+	if tags == nil {
+		tags = make([]news.Tag, 0)
+	}
+
+	c.JSON(http.StatusOK, gin.H{"status": "success", "data": gin.H{"records": tags, "meta": gin.H{
+		"latest_order": q.Filter.LatestOrder,
+		"offset":       q.Offset,
+		"limit":        q.Limit,
+	}}})
 }
 
 func (nc *newsV2Controller) GetTopics(c *gin.Context) {
@@ -290,6 +331,24 @@ func (nc *newsV2Controller) getIndexPageJobs() []job {
 			v.Name,
 			typePost,
 			news.NewQuery(news.WithFilterCategoryIDs(v.ID), news.WithLimit(1)),
+		})
+	}
+
+	// v2 categories in index page
+	for _, v := range []news.CategorySet{
+		news.World,
+		news.Humanrights,
+		news.PoliticsAndSociety,
+		news.Health,
+		news.Econ,
+		news.Culture,
+		news.Education,
+		// news.Test,
+	} {
+		jobs = append(jobs, job{
+			v.Name,
+			typePost,
+			news.NewQuery(news.WithFilterCategorySet(v.Key), news.WithLimit(1)),
 		})
 	}
 
--- a/internal/news/model.go
+++ b/internal/news/model.go
@@ -51,6 +51,7 @@ type MetaOfPost struct {
 	Title                string             `bson:"title" json:"title"`
 	Subtitle             string             `bson:"subtitle" json:"subtitle"`
 	Categories           []category         `bson:"categories" json:"categories,omitempty"`
+	CategorySet          []category_set     `bson:"category_set" json:"category_set,omitempty"`
 	PublishedDate        time.Time          `bson:"publishedDate" json:"published_date"`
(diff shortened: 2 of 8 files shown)


controllers/news_v2.go
```diff
@@ -59,6 +61,14 @@ func (nc *newsV2Controller) GetPosts(c *gin.Context) {
 		return
 	}
 
+	if q.Filter.SubcategoryID != "" && len(posts) == 0 {
```
Aylie-Chou, line 64: The logic of `category & subcategory is not consistent` is designed as given subcategory not belonged to given category. (I would also add the description in design doc)

controllers/news_v2.go
```diff
@@ -293,6 +334,24 @@ func (nc *newsV2Controller) getIndexPageJobs() []job {
...
+	// v2 categories in index page
+	for _, v := range []news.CategorySet{
+		news.World,
+		news.Humanrights,
+		news.PoliticsAndSociety,
+		news.Health,
+		news.Econ,
+		news.Culture,
+		news.Education,
+		// news.Test,
```
Aylie-Chou, line 346: Please help to remove commented code.

controllers/news_v2.go
```diff
@@ -293,6 +334,23 @@ func (nc *newsV2Controller) getIndexPageJobs() []job {
...
 
+	// v2 categories in index page
+	for _, v := range []news.CategorySet{
+		news.World,
+		news.Humanrights,
+		news.PoliticsAndSociety,
+		news.Health,
+		news.Econ,
+		news.Culture,
+		news.Education,
```
Aylie-Chou, line 345: `news.Environment` is missing

internal/news/model.go
```diff
@@ -51,6 +51,7 @@ type MetaOfPost struct {
 	Title                string             `bson:"title" json:"title"`
 	Subtitle             string             `bson:"subtitle" json:"subtitle"`
 	Categories           []category         `bson:"categories" json:"categories,omitempty"`
+	CategorySet          []category_set     `bson:"category_set" json:"category_set,omitempty"`
```
Aylie-Chou, line 54: [On 8/15 disscusion](https://www.notion.so/twreporter/design-475492f936054afd8ed4ce6e9e35f513#4c7fab40ced940d08dbae97e57efa344), return meta of post's category set would fetch the category, sucategory data from postcategories & tags table.

howar31: yes, this will be implemented in the following commit

Aylie-Chou (changes requested):
Mostly looking good !
I left few comments, please have a look. 
Let me know if there's any problem with spec or design documents.

controllers/news_v2.go
howar31: if `subcategory`'s parent is not `category`, the `posts` will be nil and the length will be zero.

internal/news/model.go
howar31: fixed

howar31:
The test cases are fixing now and will be ready in next commit.

internal/news/model.go
Aylie-Chou: In the reponse of /index-page, some articles can correctly fetch category_set data, while some cannot.
螢幕截圖 2023-01-18 上午10 33 13
螢幕截圖 2023-01-18 上午10 36 27

controllers/news_v2.go
Aylie-Chou: Yes, but the reason cause empty `posts` are more than `category & subcategory is not consistent`.
 In the case the given category_set has no posts in db, we want a valid 200 response.

Also for the readability of code, it would be great if the code logic directly means its business logic.

Aylie-Chou:
The endpont `/posts/<slug>` now returns 500 with this error :
螢幕截圖 2023-01-18 上午10 33 13

Can you reproduce this problem ?

howar31:
> The endpont `/posts/<slug>` now returns 500 with this error : 螢幕截圖 2023-01-18 上午10 33 13
> 
> Can you reproduce this problem ?

It's currently functional for me.

Get a single post - TWReporter_2023-01-18_13-48-12

I'll keep taking a look for this issue.

internal/news/model.go
howar31: It's wierd that some posts are correct while some are not.  Cause is currently unknown.  Looking into it

Aylie-Chou:
When testing with main site category page, I found `/v2/posts?category_id=xxxxx` return empty record while `/v2/posts?category_id=xxxxx&subcategory_id=zzzzzz` has mutiple records.

// result of category page
螢幕截圖 2023-01-30 下午4 49 37

// result of subcategory page
螢幕截圖 2023-01-30 下午4 49 47

internal/news/model.go
Aylie-Chou: In former discussion, the type of `category_set` would be `[{ category: <ObjectId>, subcategory: <ObjectId> }]`, and it would need another db migration to make `category_set` type correct.
Cuurent code test return records correctly with correct type of category_set 👍

internal/news/mongo.go
```diff
@@ -148,6 +169,42 @@ func (mf mongoFilter) BuildElements() []bson.E {
 					elements = append(elements, mongo.BuildElement(fieldID, id))
 				}
 			}
+		case categorySet:
+			v := fieldV.Interface().(categorySet)
+
+			if v.Category != "" && v.Subcategory != "" {
+				var categoryId interface{} = v.Category
+				objectID, err := primitive.ObjectIDFromHex(v.Category)
```
Aylie-Chou, line 177: Question: what is the error handle if `err != nil` ?

howar31: if `err != nil` means the ObjectID convertion failed.  The categoryId/subcategoryId will use raw input (which is string).  The query will use the input to query DB and return results.  (If the type of the field should be ObjectId, the query will return empty)

storage/news_v2.go
```diff
@@ -328,6 +329,46 @@ func (m *mongoStorage) GetAuthorCount(ctx context.Context, q *news.Query) (int64
...
+		// document := bson.D{}
+
+		var categoryId interface{} = q.Filter.CategorySet.Category
+		objectID, err := primitive.ObjectIDFromHex(q.Filter.CategorySet.Category)
+		if err == nil {
+			categoryId = objectID
+		}
+
+		query := bson.M{
+			"$and": []bson.M{
```
Aylie-Chou, line 344: Question: why do we need `$and` wrap ?

storage/news_v2.go
```diff
@@ -328,6 +329,46 @@ func (m *mongoStorage) GetAuthorCount(ctx context.Context, q *news.Query) (int64
...
+				{"_id": categoryId},
+				{"subcategory": q.Filter.CategorySet.Subcategory},
```
Aylie-Chou, line 346: Should we also transfer `q.Filter.CategorySet.Subcategory` type to ObjectId since we store postcategories.subcategory as array of objectId ?

controllers/news_v2.go
```diff
@@ -25,6 +25,7 @@ type newsV2Storage interface {
 	GetPostCount(context.Context, *news.Query) (int64, error)
 	GetTopicCount(context.Context, *news.Query) (int64, error)
 	GetAuthorCount(context.Context, *news.Query) (int64, error)
+	GetCategoryCount(context.Context, *news.Query) (int64, error)
```
Aylie-Chou, line 28: Question: current naming `GetCategoryCount` is a bit irrelevant to the actual content. Could we rename this function as something like `CheckCategorySetValid` and return boolean ?

storage/news_v2.go
howar31: no, it's a string array in`postcategories` collection

storage/news_v2.go
howar31: removed

controllers/news_v2.go
howar31: done

Aylie-Chou (approved):
LGTM !!

PR #496, opened 2023-01-09
fix: use twreporter-gcp context for gcp related credential

Aylie-Chou:
We move gcp releated credentials/vars into `twreporter-gcp` context.
This patch update accessing context name for accessing gcp related credentials.


howar31 (approved):
LGTM

PR #516, opened 2023-02-21
fix: getMetaOfPosts sorting issue

howar31:


storage/news_v2.go
```diff
@@ -91,11 +91,11 @@ func (m *mongoStorage) GetMetaOfPosts(ctx context.Context, q *news.Query) ([]new
...
 
 	// build aggregate stages from query
-	stages := news.BuildFilterQueryStatements(mq)
+	stages := news.BuildQueryStatements(mq)
 	// build lookup(join) stages according to required fields
 	stages = append(stages, news.BuildLookupStatements(news.LookupMetaOfPost)...)
-	// build sorting and paging stages from query
-	stages = append(stages, news.BuildSortPageQueryStatements(mq)...)
+	// build sorting stages from query again to sort grouped data
+	stages = append(stages, news.BuildSortQueryStatements(mq)...)
```
Aylie-Chou, line 98: Question: could we sort twice only on `BuildCategorySetStage` since other lookup stage would not mess up the order ?

howar31: Based on the current code structure, we won't be passing mq (query input) into BuildLookupStatement. As a result, we'll have to construct the sorting query outside of the BuildLookupStatement function.

Aylie-Chou (approved):
LGTM

PR #522, opened 2023-03-13
fix: endpoint /line-notify publish to neticrm topic

Aylie-Chou:
This patch fix [[defect] /prime/line-notify 沒接到 neticrm](https://app.asana.com/0/1203699264568808/1204010979844366/f)


howar31 (approved):
LGTM

PR #531, opened 2023-03-27
fix: /posts endpoint return 404 if post.caetgory_set is empty

Aylie-Chou:
# Related Issue
[沒有分類組合的文章，頁面會呈現 404](https://app.asana.com/0/0/1204213366816443/f)

# Root Cause
`$unwind` would drop record when target field is null ([ref](https://www.mongodb.com/docs/manual/reference/operator/aggregation/unwind/#field-path-operand))

# Solution
add `preserveNullAndEmptyArrays` option to `$unwind` to preserve records.

### DIFF
--- a/internal/mongo/document.go
+++ b/internal/mongo/document.go
@@ -60,7 +60,10 @@ func BuildCategorySetStage() []bson.D {
 	var result []bson.D
 
 	// unwind category_set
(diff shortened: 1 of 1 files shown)


howar31 (approved):
LGTM

PR #537, opened 2023-04-01
feat: Add Mailgroup, SetUser related functions

howar31:
# Issue
New function for setting user's preferences, including maillist related

# Notice
- only SetUser completed
- GetUser has table join issue
- unit test will be delievered in another commit

# Dependency

### DIFF
--- a/configs/config.go
+++ b/configs/config.go
@@ -80,6 +80,11 @@ neticrm:
     project_id: "" # gcp project id
     pub_topic: "" # pub/sub topic
     slack_webhook: "" # slack notify webhook
+mailchimp:
+    interest_ids:
+        featured: b583571955 # 報導者精選
+        behind_the_scenes: 49de5e7889 # 採訪幕後故事
+        operational_journal: 092053aaa1 # 報導者營運手記
 `)
 
 type ConfYaml struct {
@@ -94,6 +99,7 @@ type ConfYaml struct {
 	Encrypt     EncryptConfig    `yaml:"encrypt"`
 	News        NewsConfig       `yaml:"news"`
 	Neticrm     NeticrmPubConfig `yaml:"neticrm"`
+	Mailchimp   MailchimpConfig  `yaml:"mailchimp"`
 }
 
 type CorsConfig struct {
@@ -196,9 +202,13 @@ type NewsConfig struct {
 }
 
 type NeticrmPubConfig struct {
-	ProjectID      string        `yaml:"project_id"`
-	Topic          string        `yaml:"pub_topic"`
-	SlackWebhook   string        `yaml:"slack_webhook"`
+	ProjectID    string `yaml:"project_id"`
+	Topic        string `yaml:"pub_topic"`
+	SlackWebhook string `yaml:"slack_webhook"`
+}
+
+type MailchimpConfig struct {
+	InterestIDs map[string]string `yaml:"interest_ids"`
 }
 
 func init() {
@@ -283,9 +293,14 @@ func buildConf() ConfYaml {
 	conf.News.IndexPageTimeout = viper.GetDuration("news.index_page_timeout")
 	conf.News.AuthorPageTimeout = viper.GetDuration("news.author_page_timeout")
 
+	// Neticrm
 	conf.Neticrm.ProjectID = viper.GetString("neticrm.project_id")
 	conf.Neticrm.Topic = viper.GetString("neticrm.pub_topic")
 	conf.Neticrm.SlackWebhook = viper.GetString("neticrm.slack_webhook")
+
+	// Mailchimp
+	conf.Mailchimp.InterestIDs = viper.GetStringMapString("mailchimp.interest_ids")
+
 	return conf
 }
 
--- a/configs/config.go
+++ b/configs/config.go
@@ -82,9 +82,10 @@ neticrm:
     slack_webhook: "" # slack notify webhook
 mailchimp:
     interest_ids:
-        featured: b583571955 # 報導者精選
-        behind_the_scenes: 49de5e7889 # 採訪幕後故事
-        operational_journal: 092053aaa1 # 報導者營運手記
+        featured: 2f1c91a75a # 報導者精選
+        behind_the_scenes: 219df4131b # 採訪幕後故事
+        operational_journal: ca9d491549 # 報導者營運手記
+        event_updates: 0345c43d67 # 活動資訊與最新動態
 `)
 
 type ConfYaml struct {
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -1,0 +1,39 @@
+package controllers
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/twreporter/go-api/models"
+)
+
+// GetUser given userID, this func will try to get the user record, joined with users_mailgroup table, from DB
+func (mc *MembershipController) GetUser(c *gin.Context) (int, gin.H, error) {
+	userID := c.Param("userID")
+
+	user, err := mc.Storage.GetUserByID(userID)
+	if err != nil {
+		return toResponse(err)
+	}
+
+	return http.StatusOK, gin.H{"status": "ok", "record": user}, nil
+}
+
+// SetUser given userID and POST body, this func will try to create record in the related table,
+// and build the relationship between records and user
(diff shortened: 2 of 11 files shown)


models/user_preferences.go
```diff
@@ -0,0 +1,16 @@
+package models
+
+import (
+	"time"
+)
+
+type UserPreference struct {
+	Topics   []string `json:"topics"`
```
Aylie-Chou, line 8: Could we use the same naming  `ReadPreference` here for the consistency

storage/mailgroup.go
```diff
@@ -0,0 +1,23 @@
+package storage
+
+import (
+	"github.com/pkg/errors"
+)
+
+// CreateMaillistOfUser this func will accept maillist string array as input, delete all entry of the user in users_mailgroup, and insert each input entry into users_mailgroup table
+func (gs *GormStorage) CreateMaillistOfUser(uid string, maillist []string) error {
+	// delete all entry of the user in users_mailgroup
+	err := gs.db.Exec("DELETE FROM users_mailgroups WHERE user_id = ?", uid).Error
```
Aylie-Chou, line 10: question: would the syntax `?` provide the same protection as sql parameter ?

Aylie-Chou:
Could we add some test case for new added endpoints ?

storage/mailgroup.go
howar31: Yes, Exec will prepare the statement and handle the injection issue.

tests/controller_user_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/twreporter/go-api/models"
+)
+
+type UserPreferences struct {
```
Aylie-Chou, line 13: Could we use the defined `UserPreferences` model in `models/user_preferences.go` here ?

tests/controller_user_test.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+func TestSetUser_Success(t *testing.T) {
+	// Mocking user
+	var user models.User = getUser(Globs.Defaults.Account)
+	jwt := generateIDToken(user)
+
+	// Mocking preferences
+	preferences := UserPreferences{
+		ReadPreference: []string{"test1", "test2"},
```
Aylie-Chou, line 25: `read_preference` field has defined acceptable value list.
`test1`, `test2`  would not be allowed to save to db

Aylie-Chou (changes requested):
Appreciate the hard work !
We would still need a list for `maillist` from api attributes to mailchimp id.
And also it would be better to add failed case in `user_test.go`

I left some cooments, please have a look.

tests/controller_user_test.go
howar31: done

tests/controller_user_test.go
howar31: done

controllers/user.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		fmt.Println("Error decoding JSON:", err)
+	}
+
+	// Call UpdateReadPreferencetOfUser to save the preferences.ReadPreference to DB
+	if err = mc.Storage.UpdateReadPreferencetOfUser(userID, preferences.ReadPreference); err != nil {
+		return toResponse(err)
+	}
+
+	if !isValidMaillistValue(preferences.Maillist) {
```
Aylie-Chou, line 38: Could we check maillist validation before update read preference since `SetUser` would return error despite update read preference success or not.

storage/user.go
```diff
@@ -182,3 +183,20 @@ func (gs *GormStorage) UpdateReporterAccount(ra models.ReporterAccount) error {
...
+// UpdateReadPreferencetOfUser function will update the read preference of a user
+func (gs *GormStorage) UpdateReadPreferencetOfUser(userID string, readPreference []string) error {
+	user := models.User{}
+	err := gs.db.First(&user, userID).Error
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("get user(id: %s) error", userID))
+	}
+
+	user.ReadPreference = null.StringFrom(strings.Join(readPreference, ","))
+	err = gs.db.Save(&user).Error
```
Aylie-Chou, line 196: Question: why do we need to save all `user` object here (instead of saving `read_preference` field only)
If there's other function updating another field of user, this may cover other's results.

Aylie-Chou (changes requested):
The test cases look good on this version 👍
Consider that `mailchimp` also have staging environment, we would need to make interest ids configs to support it.
Please have a look to the new added comments & feel free to discuss !

storage/user.go
howar31: By saving the entire user object, we guarantee that the changes made to ReadPreference and any other fields are saved atomically. Atomicity ensures that either all changes are saved successfully or none at all. If we were to save individual fields separately, it would be harder to ensure the atomicity of updates across multiple fields.

controllers/user.go
howar31: done

models/user_preferences.go
```diff
@@ -0,0 +1,22 @@
...
+	Maillist       []string `json:"maillist"`
+}
+
+type UsersMailGroup struct {
+	UserID      int
+	MailgroupID string
+	CreatedAt   time.Time
+}
+
+var InterestIDs = []string{
```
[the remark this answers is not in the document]
howar31: done

storage/mailgroup.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/pkg/errors"
+)
+
+// CreateMaillistOfUser this func will accept maillist string array as input, delete all entry of the user in users_mailgroup, and insert each input entry into users_mailgroup table
+func (gs *GormStorage) CreateMaillistOfUser(uid string, maillist []string) error {
+	// delete all entry of the user in users_mailgroup
+	err := gs.db.Exec("DELETE FROM users_mailgroups WHERE user_id = ?", uid).Error
+	if err != nil {
+		return errors.Wrap(err, "insert user mailgroup error")
```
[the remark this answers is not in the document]
howar31: done

storage/user.go
howar31: done

models/user_preferences.go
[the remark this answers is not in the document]
Aylie-Chou: Consider that mailchimp also have staging environment, we would need to make interest ids configs to support it.

storage/user.go
Aylie-Chou: Good job to use `transaction` !
After applying transaction, do we still have to query & save whole user object ?
(we do not check the validation of queried object)

Aylie-Chou (changes requested):
The `transaction` looks good.
There's two comment left to resolve:
- the `mailchimp` interest id would need configurated
- `transaction` pormise the atomic of `UpdateReadPreferenceOfUser`.  The previous `query & save` could be simplified as:
  - query & check queried data(id) & save target field only
    - I vote this one cause userId is the user-input parameter
  - save target field only

models/user_preferences.go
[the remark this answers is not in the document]
howar31: done

configs/config.go
```diff
@@ -80,6 +80,11 @@ neticrm:
     project_id: "" # gcp project id
     pub_topic: "" # pub/sub topic
     slack_webhook: "" # slack notify webhook
+mailchimp:
+    interest_ids:
+        featured: b583571955 # 報導者精選
+        behind_the_scenes: 49de5e7889 # 採訪幕後故事
+        operational_journal: 092053aaa1 # 報導者營運手記
```
Aylie-Chou, line 87: May we set the default value of interest id as staging version

models/user_preferences.go
```diff
@@ -0,0 +1,22 @@
...
+	Maillist       []string `json:"maillist"`
+}
+
+type UsersMailGroup struct {
+	UserID      int
+	MailgroupID string
+	CreatedAt   time.Time
+}
+
+var InterestIDs = map[string]string{
```
Aylie-Chou, line 18: Would we still need this var after add interest id config ?

configs/config.go
howar31: done

Aylie-Chou (approved):
LGTM

PR #552, opened 2023-06-06
feat: implement GetUser API and activated status

howar31:
# Issue
- New API: GetUser
- New migration: activated field in users table
- Modify API: add activated in login response
- Modify API: set activated after set interests and donate step 2

# Notice

# Dependency

---
  - https://app.asana.com/0/0/1204378944314708
  - https://app.asana.com/0/0/1204692921007278

### DIFF
--- a/controllers/donation.go
+++ b/controllers/donation.go
@@ -1,8 +1,8 @@
 package controllers
 
 import (
-	"context"
 	"bytes"
+	"context"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/rand"
@@ -22,11 +22,11 @@ import (
 	"github.com/gin-gonic/gin/binding"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"github.com/twreporter/go-mod-lib/pkg/cloudpub"
+	"github.com/twreporter/go-mod-lib/pkg/slack"
 	f "github.com/twreporter/logformatter"
 	"gopkg.in/go-playground/validator.v8"
 	"gopkg.in/guregu/null.v3"
-	"github.com/twreporter/go-mod-lib/pkg/cloudpub"
-	"github.com/twreporter/go-mod-lib/pkg/slack"
 
 	"github.com/twreporter/go-api/globals"
 	"github.com/twreporter/go-api/models"
@@ -115,17 +115,17 @@ var linePayMethods = []string{
 
 type (
 	clientReq struct {
-		Amount       uint                 `json:"amount" binding:"required"`
-		Cardholder   models.Cardholder    `json:"donor" binding:"required,dive"`
-		Receipt      models.Receipt       `json:"receipt" binding:"required,dive"`
-		Currency     string               `json:"currency"`
-		Details      string               `json:"details"`
-		Frequency    string               `json:"frequency"`
-		MerchantID   string               `json:"merchant_id"`
-		PayMethod    string               `json:"pay_method"`
-		Prime        string               `json:"prime" binding:"required"`
-		UserID       uint                 `json:"user_id" binding:"required"`
-		MaxPaidTimes uint                 `json:"max_paid_times"`
+		Amount       uint              `json:"amount" binding:"required"`
+		Cardholder   models.Cardholder `json:"donor" binding:"required,dive"`
+		Receipt      models.Receipt    `json:"receipt" binding:"required,dive"`
+		Currency     string            `json:"currency"`
+		Details      string            `json:"details"`
+		Frequency    string            `json:"frequency"`
+		MerchantID   string            `json:"merchant_id"`
+		PayMethod    string            `json:"pay_method"`
+		Prime        string            `json:"prime" binding:"required"`
+		UserID       uint              `json:"user_id" binding:"required"`
+		MaxPaidTimes uint              `json:"max_paid_times"`
 	}
 
 	clientResp struct {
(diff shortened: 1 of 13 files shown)


doc/user.apib
```diff
@@ -1,13 +1,14 @@
...
-    + name: Full Name (string) - The full name of the user
-    + activated: true (boolean) - The activated status of the user (e.g. true / false)
-    + read_preference (array) - Subscribed topics
-    + maillist (array) - Subscribed mail list
++ user_id: 123 (string) - The unique identifier of the user
++ email: example@email.com (string) - The email address of the user
++ name: Full Name (string) - The full name of the user
++ roles: explorer, action_taker, trailblazer (array) - The roles of the user
++ registration_date: `2018-01-24T22:16:48Z` (string) - The registration date of the user
++ activated: `2023-06-01T01:23:45Z` (string) - The activated status of the user. If the user is not activated, the value is null.
```
Aylie-Chou, line 9: According to [design doc](https://www.notion.so/twreporter/User-API-6fb09be7357745dbbabf68d0fbc278ad?pvs=4), the type of `activated` column would be `boolean`.
Do we have any specific reason to change its type ?

storage/user.go
```diff
@@ -17,8 +17,9 @@ import (
 func (gs *GormStorage) GetUserByID(userID string) (models.User, error) {
 	user := models.User{}
 
-	// SELECT * FROM users WHERE ID = $userID
-	if err := gs.db.First(&user, "id = ?", userID).Error; err != nil {
+	// SELECT * FROM users WHERE ID = $userID and join roles table by users_roles table
+	err := gs.db.Preload("Roles").First(&user, "id = ?", userID).Error
```
Aylie-Chou, line 21: According to api doc, the return data of `user.role` would be role name string.
Looks like we could preload `Roles.name` only here ?

storage/user.go
```diff
@@ -17,8 +17,9 @@ import (
 func (gs *GormStorage) GetUserByID(userID string) (models.User, error) {
```
Aylie-Chou, line 17: As previous discussion, the return data `user.maillist` would need to be mapped to maillist keys.

doc/user.apib
```diff
@@ -1,13 +1,14 @@
...
-+ data (object, required) - The user object
-    + user_id: 123 (string) - The unique identifier of the user
-    + email: example@email.com (string) - The email address of the user
-    + name: Full Name (string) - The full name of the user
-    + activated: true (boolean) - The activated status of the user (e.g. true / false)
-    + read_preference (array) - Subscribed topics
-    + maillist (array) - Subscribed mail list
++ user_id: 123 (string) - The unique identifier of the user
++ email: example@email.com (string) - The email address of the user
++ name: Full Name (string) - The full name of the user
```
Aylie-Chou, line 6: As previous discussion, please help to change `name` into firstname, lastname.

controllers/donation.go
```diff
@@ -629,18 +629,28 @@ func (mc *MembershipController) CreateAPeriodicDonationOfAUser(c *gin.Context) (
...
+			ID:          tokenDonation.ID,
 			OrderNumber: tokenDonation.OrderNumber,
-			Type: globals.TokenDonationType,
+			Type:        globals.TokenDonationType,
 		},
 	}
 	go publishToNeticrm(ms)
 
+	// Update the user's activated time to now
+	matchedUser, err := mc.Storage.GetUserByEmail(reqBody.Cardholder.Email)
```
Aylie-Chou, line 645: Could we update activated value with concurrency in order not to block `CreateAPeriodicDonationOfAUser` ?

Aylie-Chou (changes requested):
Mostly looking good.
I left some comments about previous discussion or mismatch deisign, please have a look.

doc/user.apib
howar31: As the discussion, it was decided to change the boolean data type to timestamp in order to store more information without significantly impacting performance.

storage/user.go
howar31: We loaded `name` and `name_en` here.  I think it's ok to preload Roles.

storage/user.go
howar31: according to Asana tasks, we return email, registered date and roles.  maillist is not implemented in this commit.

doc/user.apib
howar31: done

controllers/donation.go
howar31: done

Aylie-Chou: if the return data of `GetUserByEmail` is only for updating user's activated status, could we also move it into the concurrent function ?

doc/user.apib
```diff
@@ -1,13 +1,15 @@
...
-    + email: example@email.com (string) - The email address of the user
-    + name: Full Name (string) - The full name of the user
-    + activated: true (boolean) - The activated status of the user (e.g. true / false)
-    + read_preference (array) - Subscribed topics
-    + maillist (array) - Subscribed mail list
++ user_id: 123 (string) - The unique identifier of the user
++ email: example@email.com (string) - The email address of the user
++ fisrt_name: First Name (string) - The first name of the user
++ last_name: Last Name (string) - The last name of the user
++ roles: explorer, action_taker, trailblazer (array) - The roles of the user
```
Aylie-Chou, line 8: looks like the return value of `roles` is array of object. Could we specify what info would be contianed in the object here ?

storage/user.go
Aylie-Chou: updated in commit [3285298](https://github.com/twreporter/go-api/pull/552/commits/3285298385862ca61aa94dd4e982b6e571193df2).

Aylie-Chou (changes requested):
Thanks for the quick resolving for the `getUser` return data.
Please have a look at the comments & feel free for any discussion.

controllers/donation.go
howar31: done

doc/user.apib
howar31: done

Aylie-Chou (approved):
LGTM 👍

PR #556, opened 2023-06-27
feat: add role assign flow

howar31:
# Issue
- Add role assign function
- Add default role flow
- Add donation role flow

# Notice

# Dependency


configs/constants/constants.go
```diff
@@ -26,4 +26,11 @@ const (
 	PrivilegeMember = 10
 	// PrivilegeAdmin ...
 	PrivilegeAdmin = 50
+
+	/* Role */
+	RoleIdExplorer    = 1
```
Aylie-Chou, line 31: Question: why do we choose `id` rather than `key` to be the constants ?

howar31: The `key` is a new column.  I'll merge the `key` related commit to this PR.

Aylie-Chou (changes requested):
The code looks good with the role key !
Just have to check the details when assigning roles relationship. Please have a look at the comment.

configs/constants/constants.go
howar31: done

storage/user.go
```diff
@@ -231,3 +243,24 @@ func (gs *GormStorage) UpdateReadPreferenceOfUser(userID string, readPreference
...
+	if err := gs.db.First(&role, "key = ?", roleKey).Error; err != nil {
+		return errors.Wrap(err, fmt.Sprintf("failed to find role with Key: %s", roleKey))
+	}
+
+	// Remove all existing roles from the user
+	association := gs.db.Model(&user).Association("Roles")
+	association.Clear()
+
+	// Assign the new role to the user
+	association.Append(&role)
```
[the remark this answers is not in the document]
howar31: done with Replace()

Aylie-Chou (approved):
LGTM

PR #566, opened 2023-07-14
feat: Add role assign related mails

howar31:
# Issue
Add role assign related mails:
- send_role_explorer
- send_role_actiontaker
- send_role_trailblazer
- send_role_downgrade (only route, for cronjob implemented in later task)

# Notice

# Dependency

---
  - https://app.asana.com/0/0/1204934606026369

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -124,8 +124,9 @@ func (mc *MembershipController) AuthByEmail(c *gin.Context, sendMailRoutePath st
 				return http.StatusInternalServerError, gin.H{"status": "error", "message": "Inserting new record into DB occurs error"}, nil
 			}
 		}
-
-		go mc.sendAssignRoleMail(constants.RoleExplorer, email)
+		if matchedUser.Activated.Valid && !matchedUser.Activated.Time.IsZero() {
+			go mc.sendAssignRoleMail(constants.RoleExplorer, email)
+		}
 		statusCode = http.StatusCreated
 	}
 
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -107,11 +107,16 @@ func (mc *MembershipController) AuthByEmail(c *gin.Context, sendMailRoutePath st
 
 		// try to find record by email in users table
 		matchedUser, err = mc.Storage.GetUserByEmail(email)
-		roleCheck, _ := mc.Storage.HasRole(matchedUser, constants.RoleExplorer)
+		var roleCheck bool
+		var roleCheckErr error
 		// the user record is not existed
 		if err != nil {
 			// create records both in reporter_accounts and users table
-			_, err = mc.Storage.InsertUserByReporterAccount(ra)
+			user, err := mc.Storage.InsertUserByReporterAccount(ra)
+			roleCheck, roleCheckErr = mc.Storage.HasRole(user, constants.RoleExplorer)
+			if roleCheckErr != nil {
+				log.Println("Error checking role:", roleCheckErr)
+			}
 			if err != nil {
 				return http.StatusInternalServerError, gin.H{"status": "error", "message": "Inserting new record into DB occurs error"}, nil
 			}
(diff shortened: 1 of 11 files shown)


controllers/account.go
```diff
@@ -124,6 +125,7 @@ func (mc *MembershipController) AuthByEmail(c *gin.Context, sendMailRoutePath st
 			}
 		}
 
+		go mc.sendAssignRoleMail(constants.RoleExplorer, email)
```
Aylie-Chou, line 128: Question: do we need to assign roles for the user before sending email ?

controllers/mail.go
```diff
@@ -168,6 +173,138 @@ func (contrl *MailController) SendDonationSuccessMail(c *gin.Context) (int, gin.
 	return http.StatusNoContent, gin.H{}, nil
 }
 
+func (contrl *MailController) SendRoleExplorerMail(c *gin.Context) (int, gin.H, error) {
```
Aylie-Chou, line 176: It seems that the four added functions have some common codes. Could we combine them into a function to reduce duplicate codes ?

template/role-trailblazer.tmpl
```diff
@@ -0,0 +1,822 @@
...
+
+<p dir="ltr"><span style="color:#404040"><span style="font-size:16px">★ 透過電子報了解報導幕後和我們的工作內容</span></span></p>
+
+<p dir="ltr"><span style="color:#404040"><span style="font-size:16px">★ 獲得每季基金會工作報告，了解您的捐款對《報導者》與台灣社會的價值</span></span></p>
+
+<p dir="ltr"><span style="color:#404040"><span style="font-size:16px">★ 免費參與《報導者》付費活動</span></span></p>
+
+<p dir="ltr"><span style="color:#404040"><span style="font-size:16px">★ 享有《報導者》新書出版品 6 折優惠</span></span></p>
+
+<p dir="ltr"><span style="color:#404040"><span style="font-size:16px">★ 同時成為國家兩廳院的聯名會員</span></span></p>
```
Aylie-Chou, line 623: It seems the content of html template is a bit different with the provided doc on [dirve](https://drive.google.com/drive/folders/1T1gDVHoiuB67XdgZBkzm63nyN7oQ49Bd).
Please help to check it.

Aylie-Chou (changes requested):
Mostly looking good, and there's one thing to check:
would we need to add feature toggle here since `go-api` repo do not have `dev` branch/environment ?

template/role-trailblazer.tmpl
howar31: the template was updated on 20th July after this commit :/

controllers/account.go
howar31: The role has been assigned in `InsertUserByReporterAccount` before sending the mail.

controllers/mail.go
howar31: done

Aylie-Chou (changes requested):
Previous comments all resolved !

As the discussion in refinement meeting, please help to check following feature:
- If user already on that role, we would not send upgrading mail again
- sending explorer email only when user is `activated`

Feel free for any questions/discussions !

controllers/account.go
```diff
@@ -106,6 +107,7 @@ func (mc *MembershipController) AuthByEmail(c *gin.Context, sendMailRoutePath st
 
 		// try to find record by email in users table
 		matchedUser, err = mc.Storage.GetUserByEmail(email)
+		roleCheck, _ := mc.Storage.HasRole(matchedUser, constants.RoleExplorer)
```
Aylie-Chou, line 110: Question: how would we handle the error of `HasRole` ?

controllers/donation.go
```diff
@@ -777,10 +818,15 @@ func (mc *MembershipController) CreateADonationOfAUser(c *gin.Context) (int, gin
 		}
 
 		// Call AssignRoleToUser to assign role to user
+		roleCheck, _ := mc.Storage.HasRole(matchedUser, constants.RoleActionTaker)
```
Aylie-Chou, line 821: Question: if the user has already been `trailblazer`, would the role changed to `action taker` here ?

Aylie-Chou (changes requested):
Please have a look at comments & would we add `feature toggle` in this pr ?

controllers/donation.go
howar31: Yes, it will.  The `HasRole` only check for specific role (`Action Taker` here, for example).

Aylie-Chou: Is the role downgrading the expected spec ?

controllers/account.go
Aylie-Chou: it sounds good.
could we also add the same handling in other `HasRole` check place (`controller/donation`) ?

howar31: All `HasRole` has similar error handlings.  They won't effect the original logics.

controllers/donation.go
howar31: Fix patched

howar31:
According to discussion in meeting, the feature toggle will be implemented in another PR

controllers/donation.go
howar31: orz

PR #573, opened 2023-08-09
fix: change IsTrailblazer SQL condition

howar31:
# Issue
The IsTrailblazer sometimes failed

# Notice

# Dependency


nickhsine:
LGTM 👍

PR #576, opened 2023-08-09
fix: remove unused string from template

howar31:
# Issue
The `*|MC_PREVIEW_TEXT|*`is currently not using.

# Notice

# Dependency


nickhsine:
LGTM!

PR #577, opened 2023-08-09
fix: move role mail from signin to setUser

howar31:
# Issue
The Explorer role mail should be sent if only the user is activated and not already an Explorer.
Due to activated which is set after onboarding, move role mail from signin to setUser.

# Notice

# Dependency

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -91,19 +91,20 @@ func (mc *MembershipController) SetUser(c *gin.Context) (int, gin.H, error) {
 	if err != nil {
 		return http.StatusInternalServerError, gin.H{"status": "error", "message": "user does not exist"}, err
 	}
-	roleCheck, roleCheckErr := mc.Storage.HasRole(user, constants.RoleExplorer)
-	if roleCheckErr != nil {
-		log.Println("Error checking role:", roleCheckErr)
-	}
-
-	log.WithFields(log.Fields{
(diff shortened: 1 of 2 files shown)


controllers/user.go
```diff
@@ -84,6 +86,27 @@ func (mc *MembershipController) SetUser(c *gin.Context) (int, gin.H, error) {
...
+	}
+
+	log.WithFields(log.Fields{
+		"user.Activated.Valid":         user.Activated.Valid,
+		"user.Activated.Time.IsZero()": user.Activated.Time.IsZero(),
+		"roleCheck":                    roleCheck,
+		"sendAssignRoleMail":           !user.Activated.Valid && user.Activated.Time.IsZero() && !roleCheck,
+	}).Info("SetUser Activated Role check")
+
+	if (!user.Activated.Valid || user.Activated.Time.IsZero()) && !roleCheck {
```
Aylie-Chou, line 106: Could we skip role checking function if the user has already activated ?

Aylie-Chou (changes requested):
SetUser would be called both in onboarding flow & account page.
One user could have only one time to occur onboarding & unlimited time to set user in account page. It would be great if we can call role checking & sending email functions only in onboarding part.

Aylie-Chou (approved):
LGTM

PR #580, opened 2023-08-09
fix: only send Explorer role mail if is Explorer

howar31:
# Issue
The Explorer role was assigned while user created.  So we should send role mail if it is explorer. (not blocking it)

# Notice

# Dependency

### DIFF
--- a/controllers/user.go
+++ b/controllers/user.go
@@ -94,7 +94,7 @@ func (mc *MembershipController) SetUser(c *gin.Context) (int, gin.H, error) {
 	log.WithFields(log.Fields{
 		"user.Activated.Valid":         user.Activated.Valid,
 		"user.Activated.Time.IsZero()": user.Activated.Time.IsZero(),
-		"sendAssignRoleMail":           !user.Activated.Valid && user.Activated.Time.IsZero(),
+		"sendAssignRoleMail":           !user.Activated.Valid || user.Activated.Time.IsZero(),
 	}).Info("SetUser Activated Role check")
 	if !user.Activated.Valid || user.Activated.Time.IsZero() {
 		go mc.sendAssignRoleMail(constants.RoleExplorer, user.Email.String)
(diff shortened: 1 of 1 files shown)


controllers/user.go
```diff
@@ -102,10 +102,10 @@ func (mc *MembershipController) SetUser(c *gin.Context) (int, gin.H, error) {
 			"user.Activated.Valid":         user.Activated.Valid,
 			"user.Activated.Time.IsZero()": user.Activated.Time.IsZero(),
 			"roleCheck":                    roleCheck,
-			"sendAssignRoleMail":           !user.Activated.Valid && user.Activated.Time.IsZero() && !roleCheck,
+			"sendAssignRoleMail":           !user.Activated.Valid && user.Activated.Time.IsZero() && roleCheck,
 		}).Info("SetUser Activated Role check")
 
-		if !roleCheck {
+		if roleCheck {
```
Aylie-Chou, line 108: Question: would we not need to send explorer email if the `roleCheck` fail ?

howar31: The `roleCheck` currently function as "checking the user has the Explorer role".  So I think if the user was not an Explorer, we should not send Explorer mail to the user.

Aylie-Chou: how about the error case (where `roleCheckErr` != nil) ?

Aylie-Chou: It seems that above flow may miss sending explorer mail for users (the error cases).
I think you might have to check the error handling spec with PM.

controllers/user.go
```diff
@@ -91,23 +91,13 @@ func (mc *MembershipController) SetUser(c *gin.Context) (int, gin.H, error) {
 	if err != nil {
 		return http.StatusInternalServerError, gin.H{"status": "error", "message": "user does not exist"}, err
 	}
-
+	log.WithFields(log.Fields{
+		"user.Activated.Valid":         user.Activated.Valid,
+		"user.Activated.Time.IsZero()": user.Activated.Time.IsZero(),
+		"sendAssignRoleMail":           !user.Activated.Valid && user.Activated.Time.IsZero(),
```
Aylie-Chou, line 97: && or || ?

howar31: fixed

Aylie-Chou (approved):
LGTM

PR #583, opened 2023-08-16
fix: replacement keyword

howar31:
# Issue
The string replacement keyword in Trailblazer mail is incorrect.

# Notice

# Dependency

---
  - https://app.asana.com/0/0/1205281535004071


Aylie-Chou (approved):
LGTM

PR #587, opened 2023-08-17
feat: Add feature toggle for sendRoleMail

howar31:
# Issue
Add feature toggle for sendRoleMail.

# Notice

# Dependency

### DIFF
--- a/controllers/mail.go
+++ b/controllers/mail.go
@@ -184,22 +184,22 @@ func (contrl *MailController) sendRoleMail(c *gin.Context, subject, templateName
 		return http.StatusBadRequest, gin.H{"status": "fail", "data": failData}, nil
 	}
 
-	var templateData = struct {
-		Subject     string
-		CurrentYear string
-	}{
-		subject,
-		fmt.Sprintf("%d", time.Now().Year()),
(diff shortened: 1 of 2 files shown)


controllers/mail.go
```diff
@@ -197,9 +198,14 @@ func (contrl *MailController) sendRoleMail(c *gin.Context, subject, templateName
 
 	mailBody = out.String()
 
-	// send email through mail service
-	if err = contrl.MailService.Send(reqBody.Email, subject, mailBody); err != nil {
-		return http.StatusInternalServerError, gin.H{"status": "error", "message": fmt.Sprintf("can not send role mail to %s", reqBody.Email)}, err
+	if globals.Conf.Features.EnableRolemail {
```
Aylie-Chou, line 201: Question: why do we choose to toggle here instead of the beginning of `sendRoleMail` function ?

howar31: I refine the start point of the feature toggle to include only necessary lines.

Aylie-Chou (approved):
LGTM
don't forget to update config map in kubernetes-config repo to include new added config value.

PR #606, opened 2023-09-04
feat: Add OTP login APIs

howar31:
# Issue
Add OTP login APIs
- /v1/mail/send_otp
- /v3/auth/signin
- /v3/auth/activate

# Notice

# Dependency

---
  - https://app.asana.com/0/0/1204934606026373

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"fmt"
+	"math/rand"
 	"net/http"
 	"net/mail"
 	"net/url"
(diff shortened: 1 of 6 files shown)


controllers/account.go
```diff
@@ -382,3 +383,216 @@ func (mc *MembershipController) Onboarding(c *gin.Context) (int, gin.H, error) {
...
+			Email:         email,
+			ActivateToken: otpCode,
+			// expire time is 15 minute
+			ActExpTime: time.Now().Add(time.Duration(15) * time.Minute),
+		}
+
+		// try to find record by email in users table
+		matchedUser, err = mc.Storage.GetUserByEmail(email)
+		// the user record is not existed
+		if err != nil {
```
Aylie-Chou, line 455: Question: should we determine error case or matchedUser nil case here ?

controllers/account.go
```diff
@@ -382,3 +383,216 @@ func (mc *MembershipController) Onboarding(c *gin.Context) (int, gin.H, error) {
...
+
+	var activatedString string
+	if user.Activated.Valid && !user.Activated.Time.IsZero() {
+		activatedString = user.Activated.Time.Format(time.RFC3339)
+	}
+
+	c.SetCookie("activated", activatedString, idTokenExpiration, defaultPath, defaultDomain, secure, true)
+	c.SetCookie("id_token", idToken, idTokenExpiration, defaultPath, defaultDomain, secure, true)
+
+	return http.StatusOK, gin.H{
```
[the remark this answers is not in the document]
howar31: As discussion in meeting, the redirect will be implemented by frontend.  Backend will set cookie header only.

controllers/account.go
```diff
@@ -382,3 +383,216 @@ func (mc *MembershipController) Onboarding(c *gin.Context) (int, gin.H, error) {
...
+				Email: email,
+			},
+		}, errors.New(errMsg)
+	}
+
+	// check expire time
+	if ra.ActExpTime.Before(time.Now()) {
+		errMsg := "otp code expired"
+		return http.StatusForbidden, gin.H{
+			"status": "fail",
```
Aylie-Chou, line 543: Could we use `expired` as status value here to help frontend identify the fail cases 🙏

howar31: done

Aylie-Chou (changes requested):
There's one comment left, please have a look ~
Also could we have test cases for the new added function & update `/v3/activate` route `403` error info in api doc in next PR (not blocking this feature, but also important) 🙏

controllers/account.go
howar31: According to other similar functions (L110, L133), we do not need to determine error case or matchedUser nil case here.

Aylie-Chou: If `matchedUser` and `err` are both nil, current code would still mark the user existed.
Is that the expected result ?

Aylie-Chou (commented):
Please have a look at the new added comment.

controllers/account.go
howar31: I tried the `GetUserByEmail()`.  If `matchedUser` is nil (user not found), the `err` will return `record not found` error.  Normally they won't both be nil.

Aylie-Chou: Got it !
Thanks for the checking.

Aylie-Chou (approved):
LGTM

PR #616, opened 2023-10-04
fix: isTrailblazer sql logic

howar31:
# Issue
Fix isTrailblazer sql logic from `pd` to `p`

# Notice

# Dependency


nickhsine:
LGTM

PR #621, opened 2023-10-04
feat: Add roles, activated to JWT access token

howar31:
# Issue
Add roles, activated to JWT access token

# Notice

# Dependency

---
  - https://app.asana.com/0/0/1205522078452851


controllers/account.go
```diff
@@ -298,7 +298,22 @@ func (mc *MembershipController) TokenDispatch(c *gin.Context) (int, gin.H, error
 		return http.StatusInternalServerError, gin.H{"status": "error", "message": "cannot get user data"}, err
 	}
 
-	accessToken, err = utils.RetrieveV2AccessToken(user.ID, user.Email.ValueOrZero(), acccessTokenExpiration)
+	roles := make([]map[string]interface{}, len(user.Roles))
+	for i, role := range user.Roles {
```
Aylie-Chou, line 302: Question: if the user has multiple roles, what role would we supposed to return here ?

howar31: In current implementation, all roles will be returned in an array (same as API response).  For multiple roles support, it should be implemented in another commit imo.

Aylie-Chou: As the disscusion of [this thread](https://github.com/twreporter/cronjob/pull/104#discussion_r1287880344), would we apply it in this pr ? (or we could apply the new multiple roles logic when conflict fixing in this sprint)

howar31: done.  both token and API are fixed.

Aylie-Chou (approved):
feel free to discuss if you have any problem while resolving conflict

howar31:
I've resolved the conflict by:
1. order by `roles`.`weight` DESC first
2. order by `users_roles`.`updated_at` DESC

Aylie-Chou (approved):
LGTM, we could update other getting roles function with the order `roles.weight desc, users_roles.updated_at desc` in following PR.

PR #622, opened 2023-10-04
feat: send same activeToken in 5 min for signin

howar31:
# Issue
send same activeToken in 5 min for signin

# Notice

# Dependency

---
  - https://app.asana.com/0/0/1205413135009481

### DIFF
--- a/controllers/account.go
+++ b/controllers/account.go
@@ -83,13 +83,23 @@ func (mc *MembershipController) AuthByEmail(c *gin.Context, sendMailRoutePath st
 
 	// get reporter account by email from reporter_account table
 	ra, err = mc.Storage.GetReporterAccountData(email)
+
 	// account is already signed in before
 	if err == nil {
-		// update active token and token expire time
-		ra.ActivateToken = activeToken
-		ra.ActExpTime = time.Now().Add(time.Duration(15) * time.Minute)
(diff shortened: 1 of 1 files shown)


controllers/account.go
```diff
@@ -83,13 +83,23 @@ func (mc *MembershipController) AuthByEmail(c *gin.Context, sendMailRoutePath st
...
 	if err == nil {
-		// update active token and token expire time
-		ra.ActivateToken = activeToken
-		ra.ActExpTime = time.Now().Add(time.Duration(15) * time.Minute)
-		if err = mc.Storage.UpdateReporterAccount(ra); err != nil {
-			return http.StatusInternalServerError, gin.H{"status": "error", "message": "Updating DB occurs error"}, err
+		// Calculate the time difference between ActExpTime and now
+		timeDifference := ra.ActExpTime.Sub(time.Now())
+
+		if timeDifference > 10*time.Minute {
```
Aylie-Chou, line 92: Question: is the expired time 10 minutes or 5 minutes ?

howar31: expired time is 5 minutes

Aylie-Chou: Got it, could we add a comment to idicate the spec that the activate token would not change in 5 minutes ?

howar31: done

Aylie-Chou (approved):
LGTM

PR #639, opened 2023-11-27
feat: user post reading count & time

Aylie-Chou:
# Issue
[[會員制]閱讀文章篇數、閱讀時間-後端](https://app.asana.com/0/1203699264568808/1205988685100790/f)

# Notice
- add migration files
- add api doc
- add test cases

# Dependency
N/A

### DIFF
--- a/controllers/analytics.go
+++ b/controllers/analytics.go
@@ -1,0 +1,64 @@
+package controllers
+
+import (
+	"fmt"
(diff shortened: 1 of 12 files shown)


Aylie-Chou:
The procedure of dealing with disagree data collection has not been established, so I think we could leave the api to change `agree_data_collection` value to future.

PR #652, opened 2023-12-29
fix: use new category_set key for review section

liruchen32:
# Issue
[報導者首頁「評論」區塊未顯示最新文章](https://app.asana.com/0/0/1206238664023561/f)

# Notice

# Dependency

### DIFF
--- a/controllers/news_v2.go
+++ b/controllers/news_v2.go
@@ -302,7 +302,7 @@ func (nc *newsV2Controller) getIndexPageJobs() []job {
 			Name: news.ReviewsSection,
 			Type: typePost,
 			Query: news.NewQuery(
-				news.WithFilterCategoryIDs(news.Review.ID),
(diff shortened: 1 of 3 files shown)


controllers/news_v2.go
```diff
@@ -302,7 +302,7 @@ func (nc *newsV2Controller) getIndexPageJobs() []job {
 			Name: news.ReviewsSection,
 			Type: typePost,
 			Query: news.NewQuery(
-				news.WithFilterCategoryIDs(news.Review.ID),
```
Aylie-Chou, line 305: Question: is `Review` const been used in other place ? (If not, could we remove it)

Aylie-Chou (changes requested):
Please have a look at comment ~

controllers/news_v2.go
liruchen32: fixed

Aylie-Chou (approved):
LGTM

PR #661, opened 2024-01-22
fix: get current 6 month footprints only

Aylie-Chou:
# Issue
/reading-footprint endpoint should only return current 6 month data

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #685, opened 2024-03-11
fix: query with category_set instead of categories

Aylie-Chou:
# Issue
- /index_page opinion & photography section posts wrong
- /posts query with category & without subcategory would get wrong posts

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #704, opened 2024-04-16
feat: add more card info

liruchen32:
# Issue
add card info for phase 2

# Notice

# Dependency


Aylie-Chou (changes requested):
Could we also update api doc for this change ?

Aylie-Chou (approved):
LGTM
Thanks for covering~~

PR #734, opened 2024-07-10
fix: use First to find bookmark and dont return err while not found

liruchen32:
# Issue
[[維運] redux 升級後續 refactor: 書籤](https://app.asana.com/0/1203699264568808/1207639742964814/f) defects

# Notice

# Dependency

### DIFF
--- a/storage/news_v2.go
+++ b/storage/news_v2.go
@@ -127,13 +127,19 @@ func (gs *gormStorage) GetBookmarksForFullPost(ctx context.Context, userID strin
 func (gs *gormStorage) GetBookmarksForFullPostTask(ctx context.Context, userID string, post news.Post) <-chan fetchResult {
 	result := make(chan fetchResult)
 	go func(ctx context.Context, userID string, post news.Post) {
+		defer close(result)
 
 		postID := post.ID.Hex()
 
 		var UserBookmark models.UsersBookmarks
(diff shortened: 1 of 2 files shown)


storage/news_v2.go
```diff
@@ -127,13 +127,19 @@ func (gs *gormStorage) GetBookmarksForFullPost(ctx context.Context, userID strin
 func (gs *gormStorage) GetBookmarksForFullPostTask(ctx context.Context, userID string, post news.Post) <-chan fetchResult {
 	result := make(chan fetchResult)
 	go func(ctx context.Context, userID string, post news.Post) {
+		defer close(result)
 
 		postID := post.ID.Hex()
 
 		var UserBookmark models.UsersBookmarks
-		err := gs.db.Where("user_id = ? AND post_id = ?", userID, postID).Find(&UserBookmark).Error
+		err := gs.db.Where("user_id = ? AND post_id = ?", userID, postID).First(&UserBookmark).Error
```
Aylie-Chou, line 135: Question: would user has multiple bookmark data with same post_id in `users_bookmarks` tables ?
Or the reason we change the method from `Find` to `First` to just to return single object than slices

Aylie-Chou (approved):
LGTM
just curious about the reason of changing `Find` to `First`

storage/news_v2.go
liruchen32: No, because the database checks for duplicate keys using user_id and post_id.

PR #771, opened 2024-09-26
fix: only return last 3 month followups

Aylie-Chou:
# Issue
[[debug]報導追蹤，可顯示項目超過三個月](https://app.asana.com/0/1203699264568808/1208375665334458/f)

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #781, opened 2024-10-07
fix: use `Header.Set` instead of `Header.Add` for testing

Aylie-Chou:
# Issue
address IAP isuue in [[電子收據] go-api 對接 member cms ](https://app.asana.com/0/1203699264568808/1208301216504807/f)

# Notice
ref: http package [doc](https://pkg.go.dev/net/http#Header.Set)

# Dependency
N/A


nickhsine (approved):
LGTM 👍

PR #832, opened 2025-05-21
feat: add name column to users table

howar31:
# Issue
[[會員管理系統] 新增/恢復全名欄位](https://www.notion.so/1e48dbdca9328084a3e2feba5937750d?pvs=8&n=github_linkback)
go-api 這邊需要新增對應 migration

# Notice

# Dependency


Aylie-Chou (commented):
Question: would we need to assign user.name value as last_name + first_name as well in this PR ?

howar31:
> Question: would we need to assign user.name value as last_name + first_name as well in this PR ?

No, there'll be another task in member-cms for that.

Aylie-Chou (approved):
LGTM

